PR #3, opened 2022-10-17
[RSDK-614] Create Vector3 and corresponding FFI namespace to start sp…

gvaradarajan:
…atialmath library

(virtually) a copy of this already approved PR: https://github.com/viamrobotics/viam-spatial-math/pull/1

### DIFF
--- a/src/ffi/dial_ffi.rs
+++ b/src/ffi/dial_ffi.rs
@@ -28,13 +28,13 @@ use anyhow::Result;
 
 use crate::proxy::grpc_proxy::GRPCProxy;
 
-/// The Ffi interface, returned as a pointer by init_rust_runtime. User should keep this pointer until freeing the runtime.
-pub struct Ffi {
+/// The DialFfi interface, returned as a pointer by init_rust_runtime. User should keep this pointer until freeing the runtime.
+pub struct DialFfi {
     runtime: Option<Runtime>,
     sigs: Option<Vec<oneshot::Sender<()>>>,
 }
 
-impl Drop for Ffi {
+impl Drop for DialFfi {
     fn drop(&mut self) {
         log::debug!("FFI runtime closing");
         if let Some(r) = self.runtime.take() {
@@ -43,7 +43,7 @@ impl Drop for Ffi {
     }
 }
 
-impl Ffi {
+impl DialFfi {
     fn new() -> Self {
         Self {
             runtime: Some(Runtime::new().unwrap()),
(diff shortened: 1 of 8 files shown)


src/ffi/spatialmath/vector3.rs
```diff
@@ -0,0 +1,121 @@
...
+pub extern "C" fn new_vector3(x: f64, y: f64, z: f64) -> *mut Vector3 {
+    Vector3::new(x, y, z).to_raw_pointer()
+}
+
+#[no_mangle]
+pub extern "C" fn free_vector_memory(ptr: *mut Vector3) {
+    if ptr.is_null() {
+        return;
+    }
+    unsafe {
```
stuqdog, line 16: Why are we not putting the `unsafe` in the signature instead? seems more consistent with how we do this everywhere else.

src/ffi/dial_ffi.rs
```diff
@@ -28,13 +28,13 @@ use anyhow::Result;
 
 use crate::proxy::grpc_proxy::GRPCProxy;
 
-/// The Ffi interface, returned as a pointer by init_rust_runtime. User should keep this pointer until freeing the runtime.
-pub struct Ffi {
+/// The DialFfi interface, returned as a pointer by init_rust_runtime. User should keep this pointer until freeing the runtime.
+pub struct DialFfi {
```
stuqdog, line 32: @njooma will python sdk need to make any changes when this is renamed?

src/ffi/spatialmath/vector3.rs
```diff
@@ -0,0 +1,121 @@
...
+        Box::from_raw(ptr);
```
stuqdog, line 17: What is this line supposed to be doing? It looks to me like it's creating a box but not doing anything with it, which would only free the ptr in a pretty roundabout way. Can this be more explicit? Note also that clippy is complaining about it.

src/ffi/dial_ffi.rs
gvaradarajan: oh no! I can change it back if we need

src/ffi/spatialmath/vector3.rs
gvaradarajan: yeah my bad

src/ffi/dial_ffi.rs
njooma: Nope I believe this should be ok!

stuqdog (approved):

PR #7, opened 2022-10-27
RSDK-828 - fix file handle leak

stuqdog:
Uses weak arc to ensure arcs are dropped appropriately, calls `close` on channels when freeing rust runtime in FFI context.

### DIFF
--- a/src/ffi/dial_ffi.rs
+++ b/src/ffi/dial_ffi.rs
@@ -11,15 +11,16 @@ use tokio::sync::oneshot;
 use tracing::Level;
 
 use crate::rpc::dial::{
-    CredentialsExt, DialBuilder, DialOptions, WithCredentials, WithoutCredentials,
+    CredentialsExt, DialBuilder, DialOptions, ViamChannel, WithCredentials, WithoutCredentials,
 };
 use libc::c_char;
 
 use crate::proxy;
 use hyper::Server;
 use std::ffi::{CStr, CString};
-use tower::{make::Shared, ServiceBuilder};
+use tower::{make::Shared, util::Either, ServiceBuilder};
 use tower_http::{
+    auth::AddAuthorization,
     trace::{DefaultMakeSpan, DefaultOnRequest, DefaultOnResponse, TraceLayer},
     LatencyUnit,
 };
@@ -32,6 +33,7 @@ use crate::proxy::grpc_proxy::GRPCProxy;
 pub struct DialFfi {
     runtime: Option<Runtime>,
     sigs: Option<Vec<oneshot::Sender<()>>>,
+    channels: Vec<Either<AddAuthorization<ViamChannel>, ViamChannel>>,
 }
 
 impl Drop for DialFfi {
@@ -48,6 +50,7 @@ impl DialFfi {
         Self {
             runtime: Some(Runtime::new().unwrap()),
             sigs: None,
+            channels: vec![],
         }
     }
     fn push_signal(&mut self, sig: oneshot::Sender<()>) {
@@ -172,18 +175,19 @@ pub unsafe extern "C" fn dial(
     } else {
         disable_webrtc = uri_str.contains(".local") || uri_str.contains("localhost");
     }
-    let server = match runtime.block_on(async move {
+    let (server, channel) = match runtime.block_on(async move {
         let dial = match payload {
-            Some(p) => tower::util::Either::A(
+            Some(p) => Either::A(
                 dial_with_cred(uri_str, p.to_str()?, allow_insec, disable_webrtc)?
                     .connect()
                     .await?,
             ),
             None => {
                 let c = dial_without_cred(uri_str, allow_insec, disable_webrtc)?;
-                tower::util::Either::B(c.connect().await?)
+                Either::B(c.connect().await?)
             }
         };
+        let channel = dial.clone();
         let g = GRPCProxy::new(dial, uri);
         let service = ServiceBuilder::new()
             .layer(
@@ -200,7 +204,7 @@ pub unsafe extern "C" fn dial(
         let server = Server::builder(conn)
             .http2_only(true)
             .serve(Shared::new(service));
-        Ok::<_, Box<dyn std::error::Error>>(server)
+        Ok::<_, Box<dyn std::error::Error>>((server, channel))
     }) {
         Ok(s) => s,
         Err(e) => {
@@ -208,6 +212,7 @@ pub unsafe extern "C" fn dial(
             return ptr::null_mut();
         }
     };
+    ctx.channels.append(&mut vec![channel]);
     let server = server.with_graceful_shutdown(async {
         rx.await.ok();
     });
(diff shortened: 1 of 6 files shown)


src/ffi/dial_ffi.rs
```diff
@@ -172,18 +175,19 @@ pub unsafe extern "C" fn dial(
...
                     .connect()
                     .await?,
             ),
             None => {
                 let c = dial_without_cred(uri_str, allow_insec, disable_webrtc)?;
-                tower::util::Either::B(c.connect().await?)
+                Either::B(c.connect().await?)
             }
         };
+        let channel = dial.clone();
```
penguinland, line 190: `channel` is defined here but not used again until line 207. Mild preference for defining it around line 206, assuming that doesn't break functionality (and if it would, please document why it needs to be way up here). 

```rust
...
let server = ...
let channel = dial.clone();
OK(server, channel)
```

src/ffi/dial_ffi.rs
```diff
@@ -248,14 +253,26 @@ pub extern "C" fn free_rust_runtime(rt_ptr: Option<Box<DialFfi>>) -> i32 {
...
+                .runtime
+                .as_ref()
+                .map(|rt| rt.block_on(async move { chan.close().await }))
+                .unwrap_or_default(),
         }
-        None => {}
     }
+    if let Some(sigs) = ctx.sigs.take() {
+        for sig in sigs {
+            let _ = sig.send(());
```
penguinland, line 272: Do we need the `let _ =` in the first place? What would happen if this line was just `sig.send(());`?

src/ffi/spatialmath/vector3.rs
```diff
@@ -13,42 +13,42 @@ pub unsafe extern "C" fn free_vector_memory(ptr: *mut Vector3) {
     if ptr.is_null() {
         return;
     }
-    Box::from_raw(ptr);
+    let _ = Box::from_raw(ptr);
```
penguinland, line 16: Same question here: why did this line change?

src/rpc/webrtc.rs
```diff
@@ -230,19 +230,22 @@ pub(crate) async fn new_peer_connection_for_client(
...
 
     negotiation_channel
         .on_message(Box::new(move |msg: DataChannelMessage| {
-            let pc = pc.clone();
+            let wpc = pc.clone();
             let nc = nc.clone();
             Box::pin(async move {
+                let pc = match wpc.upgrade() {
+                    Some(pc) => pc,
+                    None => return,
```
penguinland, line 247: This feels like an idiom for which there is some shorthand, but I'm not quite sure what it would be. Maybe
```rust
let pc = wpc.upgrade()?;
```
or something with `.unwrap()` or `if let`. I can't quite see it right now, but this feels like it should be 1 line instead of 3.

penguinland (commented):
I don't have the background context to give this a proper review, but I'm interested in taking a look anyway.

src/rpc/webrtc.rs
stuqdog: Yeah, some shorthand would be nice! We can't use the `?` operator though because the return type is `()`. I don't love adding an `unwrap()` if we can avoid it, and (personal opinion only) I think wrapping the following code block in the `if let` is a bit uglier.

src/ffi/dial_ffi.rs
stuqdog: clippy complains if we don't use the return value at all. doesn't need to be there (everything still compiles ofc), but I personally like this look better. probably best would be `let _: Result<(), ()> = sig.send(());` so we're forced to consider any type changes.

src/ffi/spatialmath/vector3.rs
stuqdog: see above re: clippy.

src/ffi/dial_ffi.rs
```diff
@@ -200,14 +204,15 @@ pub unsafe extern "C" fn dial(
...
-        Ok::<_, Box<dyn std::error::Error>>(server)
+        Ok::<_, Box<dyn std::error::Error>>((server, channel))
     }) {
         Ok(s) => s,
         Err(e) => {
             println!("Error building GRPC proxy reason : {e:?}");
             return ptr::null_mut();
         }
     };
+    ctx.channels.append(&mut vec![channel]);
```
npmenard, line 215: wouldn't push be better?

src/rpc/client_channel.rs
```diff
@@ -27,7 +30,35 @@ pub struct WebRTCClientChannel {
...
+    fn drop(&mut self) {
+        let bc = self.base_channel.clone();
+        if !bc.is_closed() {
+            let _ = tokio::spawn(async move {
+                if let Err(e) = bc.close().await {
+                    log::error!("Error closing base channel: {e}");
+                }
+            });
+        };
+        log::info!("Dropping client channel {:?}", &self);
```
npmenard, line 52: can we change if to debug!

src/rpc/client_channel.rs
```diff
@@ -42,17 +73,25 @@ impl WebRTCClientChannel {
...
+                            return;
+                        }
+                    };
                     if let Err(e) = channel.on_channel_message(msg).await {
                         log::error!("error deserializing message: {e}");
                     }
                 })
             }))
             .await;
+        log::info!("Client channel created");
```
npmenard, line 94: same can we change this to degug!?

src/ffi/dial_ffi.rs
stuqdog: channel can't be defined lower because the immediate following line (`let g = GRPCProxy::new(dial, uri);`) takes ownership of the dial.

src/ffi/dial_ffi.rs
stuqdog: 🤦 yes it definitely would. changing.

src/rpc/client_channel.rs
stuqdog: sure thing, done

src/rpc/client_channel.rs
stuqdog: sure thing, done

src/ffi/dial_ffi.rs
```diff
@@ -248,14 +253,26 @@ pub extern "C" fn free_rust_runtime(rt_ptr: Option<Box<DialFfi>>) -> i32 {
...
+            ViamChannel::Direct(_) => (),
+            ViamChannel::WebRTC(chan) => ctx
+                .runtime
+                .as_ref()
+                .map(|rt| rt.block_on(async move { chan.close().await }))
+                .unwrap_or_default(),
         }
-        None => {}
     }
+    if let Some(sigs) = ctx.sigs.take() {
```
npmenard, line 270: Wouldn't ending the signals before closing the channels better?

npmenard (approved):
LGTM just a couple of comments

src/ffi/dial_ffi.rs
stuqdog: yep it totally would, thanks

src/ffi/dial_ffi.rs
stuqdog: actually wait I think we can make this cleaner. just pushed, lemme know what you think

penguinland: Yes, I like this more! Thanks

PR #10, opened 2022-11-03
[RSDK-732] Create Quaternion and corresponding FFI namespace

gvaradarajan:


src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+        let k = (roll_cos * pitch_cos * yaw_sin) - (roll_sin * pitch_sin * yaw_cos);
+
+        Quaternion::new(real, i, j, k)
+    }
+
+    /// Converts a quaternion into euler angles. The euler angles are 
+    /// represented according to the Tait-Bryan formalism and applied 
+    /// in the Z-Y'-X" order (where Z -> yaw, Y -> pitch, X -> roll). 
+    /// The return value is a list of [roll, pitch, yaw]
+    pub fn to_euler_angles(&self) -> [f64;3] {
```
biotinker, line 56: It may be worthwhile to allow the user to specify the rotation order for to/from euler angles.

https://github.com/go-gl/mathgl/blob/v1.0.0/mgl64/quat.go#L265

src/ffi/spatialmath/mod.rs
```diff
@@ -1 +1,2 @@
-pub mod vector3;
\ No newline at end of file
+pub mod vector3;
+pub mod quaternion;
```
penguinland, line 2: I worry that I'm sounding like a broken record at this point, but I still think these files should all end in newlines, because not ending in a new line is not POSIX-compliant and some text editors aren't going to like that. I'd still be happy to help you configure your text editor to do this automatically, if you want. If you want to do it on your own, I think a decent starting point is https://thoughtbot.com/blog/no-newline-at-end-of-file

src/spatialmath/vector3.rs
```diff
@@ -111,7 +111,6 @@ mod tests {
         assert_eq!(vector.x, 1.0);
         assert_eq!(vector.y, 1.0);
         assert_eq!(vector.z, 1.0);
-        assert!(!vector.is_normalized())
```
penguinland, line 114: If this line is removed, the test it's in doesn't test anything. Was this on purpose?

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+    pub real: f64,
+    pub i: f64,
+    pub j: f64,
+    pub k: f64,
+}
+
+impl Quaternion {
+    /// Initializes a quaternion from a real component and an
+    /// imaginary 3-vector
+    pub fn new_from_vector(real: f64, imag: Vector3) -> Self {
```
penguinland, line 22: This is hopefully my own naiveté, but is this a common way to initialize a quaternion? Why would 3 of the values already be in a vector?

src/spatialmath/quaternion.rs
penguinland: Specifying the rotation order is just a permutation at the very end (or very beginning), yeah? That sounds like a nice-to-have, but not something for the MVP. I wouldn't write anything more than a TODO for it yet (but am open to being wrong!).

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+            pitch = (std::f64::consts::PI / 2.0).copysign(pitch_sin);
+            roll = (2.0 * quat.i.atan2(quat.real)) + yaw.copysign(pitch_sin);
+        } else {
+            pitch = pitch_sin.asin();
+            let roll_sin_pitch_cos = 2.0 * ((quat.real * quat.i) + (quat.j * quat.k));
+            let roll_cos_pitch_cos = 1.0 - 2.0 * ((quat.i * quat.i) + (quat.j * quat.j));
+            roll = roll_sin_pitch_cos.atan2(roll_cos_pitch_cos);
+        }
+
+
```
penguinland, line 81: Mild preference to get rid of one of these blank lines. One blank line is enough inside a function.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+        // calculate pitch and roll
+        let pitch_sin = 2.0 * ((quat.real * quat.j) - (quat.k * quat.i));
+        let pitch: f64;
+        let roll: f64;
+        // for a pitch that is π / 2, we experience gimbal lock
+        // and must calculate roll based on the real rotation and yaw
+        if pitch_sin.abs() >= 1.0 {
+            pitch = (std::f64::consts::PI / 2.0).copysign(pitch_sin);
+            roll = (2.0 * quat.i.atan2(quat.real)) + yaw.copysign(pitch_sin);
```
penguinland, line 73: I admit I haven't checked the math myself, but [Wikipedia's reference implementation](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Quaternion_to_Euler_angles_conversion) suggests that the roll is calculated the same way regardless of whether we've got gimbal lock. Are you sure about this one?

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    pub fn get_scaled(&self, factor: f64) -> Self {
+        let mut copy_quat = self.clone();
+        copy_quat.scale(factor);
+        copy_quat
+    }
+
+    pub fn conjugate(&self) -> Self {
+        Self { real: self.real, i: self.i * -1.0, j: self.j * -1.0, k: self.k * -1.0 }
```
penguinland, line 135: DIscussion question: I had expected `-1.0 * self.i`, rather than `self.i * -1.0`, because normally when you write math, the numbers come before the letters. Do other folks prefer this ordering instead? I'm not against it, but it did surprise me a little.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /// Allocates the vector to the heap with a stable memory address and
+    /// returns the raw pointer (for use by the FFI interface)
+    pub(crate) fn to_raw_pointer(&self) -> *mut Self {
```
penguinland, line 140: Naive question: does this have to be here, or could we put it in the FFI stuff instead with the code it goes with, so we don't need the comment? Maybe there's a reason it needs to be here, but I don't see it yet.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+        Box::into_raw(Box::new(*self))
+    }
+}
+
+impl std::ops::Add<Quaternion> for Quaternion {
+    type Output = Quaternion;
+
+    fn add(self, _rhs: Quaternion) -> Quaternion {
```
penguinland, line 148: Normally, having a variable start with an underscore in Rust means "suppress the compiler warnings about unused variables. We know this one isn't used yet." but this one _is_ used. Can we rename it `rhs`? Same question for several more ops below, too.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+        let quat = Quaternion::new_from_vector(1.0, imag);
+        assert_eq!(quat, expected_quat);
+    }
+
+    #[test]
+    fn quaternion_normalizes_successfully() {
+        let mut quat = Quaternion::new(
+            0.0,
+            (1.0_f64 / 3.0_f64).sqrt() * 0.5, 
+            (1.0_f64 / 3.0_f64).sqrt() * -0.5,
```
penguinland, line 250: I'm mildly surprised that we're multiplying on the right by a constant, and a negative one at that. Could these lines become `-0.5 * (1.0_f64 / 3.0_f64).sqrt())` and similar? Maybe that's too nitpicky and this is good as-is.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+        let mut quat = Quaternion::new(0.1, 0.2, 0.3, 0.4);
+        quat.scale(2.0);
+        let expected_quat = Quaternion::new(0.2, 0.4, 0.6, 0.8);
+        assert!(approx_eq!(Quaternion, quat, expected_quat));
+    }
+
+    #[test]
+    fn get_scaled_returns_scaled_quaternion() {
+        let quat = Quaternion::new(0.1, 0.2, 0.3, 0.4);
+        let quat_orig = Quaternion::new(0.1, 0.2, 0.3, 0.4);
```
penguinland, line 312: Mild preference for `let quat_orig = quat.clone()` here. By defining it twice, it feels like we don't have a single point of truth on this thing.

src/spatialmath/quaternion.rs
gvaradarajan: we [don't currently though](https://github.com/viamrobotics/rdk/blob/main/spatialmath/quaternion.go#L49), is there a reason for that?

penguinland (commented):
I haven't looked at the FFI file yet, but here's a start. Looks very good so far!

src/spatialmath/quaternion.rs
gvaradarajan: any chance we can leave this as a follow-up? I'm pretty sure the most popular representation is ZYX

src/spatialmath/vector3.rs
gvaradarajan: no, it wasn't, this was a mistake

src/spatialmath/quaternion.rs
penguinland: Gautham set me straight: the Wikipedia page is wrong/misleading! :flushed: The next paragraph down from the code I linked is about "singularities" where that implementation breaks down. Gautham has instead showed my [this page](http://marc-b-reynolds.github.io/math/2017/04/18/TaitEuler.html), which is where he actually got his implementation from.

src/spatialmath/quaternion.rs
gvaradarajan: sometimes when you've converted it from an axis angle (with vector and theta) you scale the axis angle vector by sin(theta/2), so you might want to keep it as a vector then? I have a follow-up PR that uses is this but I'm not attached to the convenience so I'm willing to remove

src/spatialmath/quaternion.rs
gvaradarajan: I can move, l'll do it for vector as well

src/spatialmath/quaternion.rs
gvaradarajan: I don't know why I made this an _, I can remove

src/spatialmath/vector3.rs
gvaradarajan: oh it wasn't a mistake, this doesn't belong here. It's tested later on

src/spatialmath/quaternion.rs
biotinker: The reason we don't currently is because the library I linked exists and does it :)
The other two common (at least, I've seen them be in use) representations are XYZ and ZYZ.
But I don't think it needs to gate this PR, as long as it goes on a to-do list (or a library that already implements this is found)

src/ffi/spatialmath/mod.rs
biotinker: Not a broken record; not ending in a newline is AFAIK explicitly against our internal style guide as well, and will break all sorts of things. Please make sure you keep calling this out wherever you see it.

src/ffi/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,127 @@
...
+    Box::from_raw(ptr);
+}
+
+/// Get the components of a quaternion as a list of C doubles, the order of the
+/// components will be (real, i, j, k)
+#[no_mangle]
+pub unsafe extern "C" fn quaternion_get_components(quat_ptr: *const Quaternion) -> *const c_double {
+    null_pointer_check!(quat_ptr);
+    let quat = *quat_ptr;
+    let components: [c_double;4] = [quat.real, quat.i, quat.j, quat.k];
```
randhid, line 44: [q] Is this output in a double to be able to talk to rdk/ another library? Ie, is this the over the wire message at the end of the day?

src/ffi/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,127 @@
...
+/// (stored at vec_ptr). The convention is x -> i, y -> j, z -> k
+#[no_mangle]
+pub unsafe extern "C" fn quaternion_set_imag_from_vector(quat_ptr: *mut Quaternion, vec_ptr: *const Vector3) {
+    null_pointer_check!(quat_ptr);
+    null_pointer_check!(vec_ptr);
+    let mut quat = *quat_ptr;
+    let imag = *vec_ptr;
+    quat.set_imag_from_vector(imag);
+}
+
```
randhid, line 95: [nit] Is there a need/want for a set_quaternion function? I  noticed you have a get for all four elements, a set for each real and imaginary component, and a set for imaginary only. Wanted to point it out, I think it's fine/less error prone as is.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    pub fn scale(&mut self, factor: f64) {
+        self.real = self.real * factor;
+        self.i = self.i * factor;
+        self.j = self.j * factor;
+        self.k = self.k * factor;
+    }
+
+    pub fn get_scaled(&self, factor: f64) -> Self {
```
randhid, line 128: am I correct in understanding the the `get` functions always apply the method on the copy of the quaternion? Is this idiomatic rust or a preference.

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    #[test]
+    fn conjugate_works() {
+        let quat = Quaternion::new(0.1, 0.2, 0.3, 0.4);
+        let expected_quat = Quaternion::new(0.1, -0.2, -0.3, -0.4);
+        assert!(approx_eq!(Quaternion, quat.conjugate(), expected_quat));
+    }
+
+    #[test]
+    fn quaternion_initializes_from_euler_angles() {
```
randhid, line 327: [nit] Check near gimbal lock cases? https://danceswithcode.net/engineeringnotes/rotations_in_3d/demo3D/rotations_in_3d_tool.html

randhid (changes requested):

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,371 @@
...
+    #[test]
+    fn set_imag_from_vector_works() {
+        let imag = Vector3::new(0.0, 0.5, 1.0);
+        let expected_quat = Quaternion::new(1.0, 0.0, 0.5, 1.0);
+        let quat = Quaternion::new_from_vector(1.0, imag);
+        assert_eq!(quat, expected_quat);
+    }
+
+    #[test]
+    fn quaternion_normalizes_successfully() {
```
randhid, line 246: A resource for a metric for this in the future: https://nfrechette.github.io/2017/08/30/math_accuracy_quat/

src/spatialmath/quaternion.rs
penguinland: I'm not aware of this being idiomatic, but if you've got 2 functions that do the same thing except one modifies the original values and the other returns new values, it makes perfect sense to me that one of the functions should just call the other under the hood. and then I don't have a preference between which one should call the other.

src/ffi/spatialmath/quaternion.rs
gvaradarajan: yes

src/spatialmath/quaternion.rs
gvaradarajan: yes. I'm not aware of it being idiomatic rust, it's just a convention I'm using

src/spatialmath/quaternion.rs
gvaradarajan: I think one of these cases is near the gimbal lock? I'll double-check

src/spatialmath/quaternion.rs
gvaradarajan: This is so cool, I'm gonna use the fast square root method instead!

src/ffi/spatialmath/quaternion.rs
gvaradarajan: yeah I'll make a set_from_components

src/spatialmath/quaternion.rs
gvaradarajan: Turns out, Rust's own sqrt and recip functions are good enough, but this did give me the idea to add a test case that the length of a quaternion after being normalized is never _greater_ than 1.0

src/spatialmath/quaternion.rs
randhid: Gimbal lock is better tested in a set of rotations. I.e. after an axis aligns with another, do we lose a degree of freedom. It doesn't occur in one rotation. If out of scope for this ad you're limiting pitch in tait-bryan to pi/2, test that instead.

src/ffi/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Free memory at the address of the quaternion pointer. Outer processes
+/// that work with Quaternions via the FFI interface MUST remember 
+/// to call this function when finished with a quaternion
+#[no_mangle]
+pub unsafe extern "C" fn free_quaternion_memory(ptr: *mut Quaternion) {
+    if ptr.is_null() {
+        return;
+    }
+    Box::from_raw(ptr);
```
npmenard, line 41: preferred let _ =  Box::from_raw(ptr);

src/ffi/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,147 @@
...
+    to_raw_pointer(&quat)
+}
+
+/// Converts a quaternion into euler angles. The euler angles are 
+/// represented according to the Tait-Bryan formalism and applied 
+/// in the Z-Y'-X" order (where Z -> yaw, Y -> pitch, X -> roll). 
+/// The return value is a pointer to a list of [roll, pitch, yaw]
+/// as C doubles
+#[no_mangle]
+pub unsafe extern "C" fn quaternion_to_euler_angles(quat_ptr: *const Quaternion) -> *const c_double {
```
npmenard, line 142: is thera a function to free this?

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub i: f64,
+    pub j: f64,
+    pub k: f64,
+}
+
+impl Quaternion {
+    /// Initializes a quaternion from a real component and an
+    /// imaginary 3-vector
+    pub fn new_from_vector(real: f64, imag: Vector3) -> Self {
+        Self{real: real, i: imag.x, j: imag.y, k: imag.z}
```
npmenard, line 22: just real

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Returns a normalized copy of the quaternion
+    pub fn get_normalized(&self) -> Self {
+        let mut copy_quat = self.clone();
+        copy_quat.normalize();
+        copy_quat
+    }
+
+    pub fn scale(&mut self, factor: f64) {
+        self.real = self.real * factor;
```
npmenard, line 121: *= would be nicer

src/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Normalizes a quaternion
+    pub fn normalize(&mut self) {
+        // let inv_sq_dp = fast_inv_sqrt(self.norm2());
+        self.scale(self.norm2().sqrt().recip())
+    }
+
+    /// Returns a normalized copy of the quaternion
+    pub fn get_normalized(&self) -> Self {
+        let mut copy_quat = self.clone();
```
npmenard, line 115: you could de reference self and no need a clone

src/ffi/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,147 @@
...
+    quat.set_imag_from_vector(imag);
+}
+
+/// Copies the imaginary components to a 3-vector (using x -> i, y -> j
+/// z -> k) and returns a pointer to the memory address of the resulting
+/// vector
+#[no_mangle]
+pub unsafe extern "C" fn quaternion_get_imaginary_vector(quat_ptr: *const Quaternion) -> *mut Vector3 {
+    null_pointer_check!(quat_ptr);
+    let quat = *quat_ptr;
```
npmenard, line 122: here and in some other places you are copying the quaternion (let quat = *quat_ptr) and then getting the image is it on purpose?

npmenard (changes requested):
Looks great!! Done with first pass.

You should run `cargo clippy` and fix the errors it surface (could be done in a 2nd PR)

Euler angles are in radians but it's not specified either in the doc or the function

All the unsafe functions should have a #safety section especially since you are working with pointers.

Cloning vs Copy, you have to decide which one you're using mixing them is a bit weird especially since they essentially do the same. I would stick with copying since its only 3/4 f64 !

src/ffi/spatialmath/quaternion.rs
gvaradarajan: to free the quaternion? Or the euler angles?

src/spatialmath/quaternion.rs
penguinland: I think that would cause side effects on `self`, whereas this function keeps `self` unchanged and returns a new object of type `Self` instead. I think it needs to stay the way it currently is.

gvaradarajan: so I tried it out and @npmenard is correct, there are no side effects because the copy constructor is called

src/ffi/spatialmath/quaternion.rs
```diff
@@ -0,0 +1,147 @@
...
+use libc::c_double;
+
+use crate::spatialmath::{quaternion::Quaternion, vector3::Vector3};
+
+/// The FFI interface for Quaternion functions and initialization. All functions are
+/// meant to be called externally from other languages
+
+/// Allocates the quaternion to the heap with a stable memory address and
+/// returns the raw pointer (for use by the FFI interface)
+fn to_raw_pointer(quat: &Quaternion) -> *mut Quaternion {
```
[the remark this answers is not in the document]
gvaradarajan: yeah this is definitely fair, but you asking this brings up something I wish I thought about earlier: should I have done it this way? Should I be copying the quaternion before returning it? I think it's still fine, but I'm not sure

src/spatialmath/quaternion.rs
penguinland: :eyes: My mistake!

src/ffi/spatialmath/quaternion.rs
gvaradarajan: it's not, it's a mistake

src/ffi/spatialmath/quaternion.rs
npmenard: to free the *const c_double, I couldn't find the proper fucntion

src/ffi/spatialmath/quaternion.rs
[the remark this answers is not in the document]
npmenard: I think it's okay the only type of scenario where I see this being an issue is if you own a reference in some rust data structure and use that reference to change the quaternion you sent to FFI using to_raw_pointer in this case the code calling the FFI won't see the changes made on the Rust side. Given that we use simple data structures we should not run into this so ok just keep this in mind as we grow the library

npmenard (approved):
LGTM

randhid (approved):

PR #13, opened 2022-11-21
[RSDK-732] remove existing code in favor of increased reliance on nal…

gvaradarajan:
…gebra crate


src/ffi/spatialmath/quaternion.rs
```diff
@@ -216,10 +222,10 @@ pub unsafe extern "C" fn normalize_quaternion(quat_ptr: *mut Quaternion) {
 /// The caller must remember to free the quaternion memory of 
 /// *both* the input and output quaternions when finished with them 
 /// using the free_quaternion_memory FFI function
-#[no_mangle]
-pub unsafe extern "C" fn quaternion_get_normalized(quat_ptr: *const Quaternion) -> *mut Quaternion {
+// #[no_mangle]
```
stuqdog, line 225: why are we commenting out the `#[no_mangle]` here? that seems intuitively wrong to me, but even if I'm mistaken then we should just delete it rather than commenting it out.

src/ffi/spatialmath/quaternion.rs
```diff
@@ -278,10 +352,12 @@ pub unsafe extern "C" fn scale_quaternion(quat_ptr: *mut Quaternion, factor: f64
 /// The caller must remember to free the quaternion memory of 
 /// *both* the input and output quaternions when finished with them 
 /// using the free_quaternion_memory FFI function
-#[no_mangle]
-pub unsafe extern "C" fn quaternion_get_scaled(quat_ptr: *const Quaternion, factor: f64) -> *mut Quaternion {
+// #[no_mangle]
```
stuqdog, line 355: see above comment re commenting out `no_mangle`

src/spatialmath/utils.rs
```diff
@@ -0,0 +1,62 @@
+use nalgebra::Quaternion;
+
+pub fn to_euler_angles(quat: Quaternion<f64>) -> [f64;3] {
```
stuqdog, line 3: I don't love returning an array here as opposed to creating a `euler_angles` struct with `roll`, `pitch`, and `yaw` as explicitly named fields. Can we fix in this pr? I understand if this is a hassle to do immediately but I think long term we should really be more precise with our type signatures, so at a minimum we should create a new ticket and add a TODO comment.

src/ffi/spatialmath/quaternion.rs
gvaradarajan: It was a mistake

src/spatialmath/utils.rs
gvaradarajan: I can do it in this PR, I assume you want me to do this with axis angles as well then?

stuqdog: D'oh, missed the `axis_angles` case, yes doing it there too would be great thank you!

src/ffi/spatialmath/axis_angle.rs
```diff
@@ -0,0 +1,69 @@
...
+/// returns the raw pointer (for use by the FFI interface)
+fn to_raw_pointer(aa: &AxisAngle) -> *mut AxisAngle {
+    Box::into_raw(Box::new(*aa))
+}
+
+/// Free memory at the address of the axis angle pointer. Outer processes
+/// that work with axis angles via the FFI interface MUST remember 
+/// to call this function when finished with an axis angle instance
+/// 
+/// # Safety
```
stuqdog, line 16: do we need an actual safety comment here? Should the line about remembering to call this be in the Safety section?

src/ffi/spatialmath/axis_angle.rs
```diff
@@ -0,0 +1,69 @@
...
+pub unsafe extern "C" fn axis_angle_from_quaternion(
+    quat: *const Quaternion<f64>
+) -> *mut AxisAngle {
+    null_pointer_check!(quat);
+    let unit_quat = UnitQuaternion::from_quaternion(*quat);
+    let axis_opt = unit_quat.axis();
+    let angle = unit_quat.angle();
+    let axis_angle = match axis_opt {
+        Some(value) => {
+            // let axis_angle: [f64; 4] = [value[0], value[1], value[2], angle];
```
stuqdog, line 59: probably should delete this commented code

src/ffi/spatialmath/axis_angle.rs
```diff
@@ -0,0 +1,69 @@
...
+            // return Box::into_raw(Box::new(axis_angle)) as *mut _
+            AxisAngle::new(value[0], value[1], value[2], angle)
+        },
+        None => {
+            // return Box::into_raw(Box::new([0.0, 0.0, 0.0, 0.0])) as *mut _
```
stuqdog, line 64: probably should delete this commented code

src/ffi/spatialmath/euler_angles.rs
```diff
@@ -0,0 +1,55 @@
...
+/// returns the raw pointer (for use by the FFI interface)
+fn to_raw_pointer(ea: &EulerAngles) -> *mut EulerAngles {
+    Box::into_raw(Box::new(*ea))
+}
+
+/// Free memory at the address of the euler angles pointer. Outer processes
+/// that work with EulerAngles via the FFI interface MUST remember 
+/// to call this function when finished with a euler angles instance
+/// 
+/// # Safety
```
stuqdog, line 16: Should the "MUST remember to call" part go in the safety section?

src/ffi/spatialmath/quaternion.rs
```diff
@@ -50,20 +52,7 @@ pub unsafe extern "C" fn new_quaternion_from_vector(
 /// 
 /// # Safety
```
stuqdog, line 53: Should we add the "MUST remember to call" line to the Safety section?

src/ffi/spatialmath/axis_angle.rs
gvaradarajan: I'll put it in both places

gvaradarajan: oh I misunderstood what you meant, yeah will change

PR #20, opened 2022-12-05
[RSDK-732] orientation vector and rotation matrices

gvaradarajan:
this will be the last PR for the spatialmath library until Jan, at which point we might start importing it into the SDKs


src/spatialmath/utils.rs
```diff
@@ -48,22 +51,306 @@ impl EulerAngles {
...
+    pub theta: f64
+}
+
+impl OrientationVector {
+    pub fn new(o_x: f64, o_y: f64, o_z: f64, theta: f64) -> Self {
+        let o_vector = UnitVector3::new_normalize(Vector3::new(o_x, o_y, o_z));
+        OrientationVector{o_vector, theta}
+    }
+
+    pub fn from_quaternion(quat: &Quaternion<f64>) -> Self {
```
npmenard, line 119: could you also implement the To and From traits?

npmenard (commented):
Done with first pass I am testing the FFI a bit on the side :). Is there any plans to generate bindings for at least C (header file with exported function etc...)?

src/spatialmath/utils.rs
gvaradarajan: I can do it for the custom structs I created, but for quaternion and rotation matrix I would have to create wrappers. Is it fine to skip those two?

src/spatialmath/utils.rs
```diff
@@ -45,42 +48,394 @@ impl EulerAngles {
...
+    fn from(quat: Quaternion<f64>) -> Self {
+        let x_quat = Quaternion::new(0.0, -1.0, 0.0, 0.0);
+        let z_quat = Quaternion::new(0.0, 0.0, 0.0, 1.0);
+
+        let conj = quat.conjugate();
+        let new_x = (quat * x_quat) * conj;
+        let new_z = (quat * z_quat) * conj;
+
+        let o_vector = UnitVector3::new_normalize(new_z.imag());
+        let theta = orientation_vector_theta_from_rotated_axes(new_x, new_z);
```
randhid, line 215: [q] Maybe this is my Rust being rusty, Why is the theta logic not written out here and bundled intoa different function? Interested in the reasoning behind the choice. 

They are bundled up together in rdk. I'm just wondering if accessing the theta part of creating the orientation vector could be a problem for users using this library to convert. Can they get to this function on it's own, and is it useful purely on it's own.

src/ffi/spatialmath/orientation_vector.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Like an axis angle, the format involves a vector axis and a rotation
+/// theta (in radians). However, unlike an axis-angle, an orientation vector alters 
+/// the axes of the given frame of reference by rotating the z-axis to the 
+/// vector axis provided. The new x-axis is the vector that is both orthogonal to 
+/// the vector axis provided anc co-planar with both the old
+/// z-axis and the vector axis (this leaves two choices for the y-axis, 
+/// but the canonical "right-hand rule" is used to select one consistently). Then, 
+/// a clockwise-rotation of theta is applied about the new-z axis
+/// 
+/// It is highly recommended not to attempt any mathematics with the orientation
```
biotinker, line 16: Euler angles should also have this warning

src/ffi/spatialmath/orientation_vector.rs
```diff
@@ -0,0 +1,70 @@
...
+use nalgebra::Quaternion;
+
+use crate::spatialmath::utils::OrientationVector;
+
+/// The FFI Interface for initialization of Viam's Orientation Vector format.
+/// Like an axis angle, the format involves a vector axis and a rotation
+/// theta (in radians). However, unlike an axis-angle, an orientation vector alters 
+/// the axes of the given frame of reference by rotating the z-axis to the 
+/// vector axis provided. The new x-axis is the vector that is both orthogonal to 
+/// the vector axis provided anc co-planar with both the old
```
biotinker, line 11: anc -> and

randhid (changes requested):

src/spatialmath/utils.rs
```diff
@@ -45,42 +48,394 @@ impl EulerAngles {
...
+            (lat / 2.0).cos(),
+            (self.theta / 2.0).cos()
+        ];
+
+        let real = c[0]*c[1]*c[2] - s[0]*c[1]*s[2];
+        let i = c[0]*s[1]*s[2] - s[0]*s[1]*c[2];
+        let j = c[0]*s[1]*c[2] + s[0]*s[1]*s[2];
+        let k = s[0]*c[1]*c[2] + c[0]*c[1]*s[2];
+
+        Quaternion::new(real, i, j, k)
```
biotinker, line 191: Not familiar with Rust, but would it be faster to put this math directly in the `new()` call and not initialize the variables?

biotinker (approved):
I don't know Rust so I'm not going to be super useful regarding the actual code, but based on descriptions of how things work and my understanding of how Rust probably works for the math, this looks good.

src/spatialmath/utils.rs
```diff
@@ -45,42 +48,394 @@ impl EulerAngles {
...
+
+        // calculate pitch and roll
+        let pitch_sin: f64 = 2.0 * ((norm_quat.w * norm_quat.j) - (norm_quat.k * norm_quat.i));
+        let pitch: f64;
+        let roll: f64;
+        // for a pitch that is π / 2, we experience gimbal lock
+        // and must calculate roll based on the real rotation and yaw
+        if pitch_sin.abs() >= 1.0 {
+            pitch = (std::f64::consts::PI / 2.0).copysign(pitch_sin);
+            roll = (2.0 * norm_quat.i.atan2(norm_quat.w)) + yaw.copysign(pitch_sin);
```
randhid, line 69: Note: We should make this change in rdk as well.

src/spatialmath/utils.rs
```diff
@@ -61,6 +91,22 @@ impl AxisAngle {
...
+    fn from(quat: Quaternion<f64>) -> Self {
+        let unit_quat = UnitQuaternion::from_quaternion(quat);
+        let axis_opt = unit_quat.axis();
+        let angle = unit_quat.angle();
+        match axis_opt {
+            Some(value) => {
+                AxisAngle::new(value[0], value[1], value[2], angle)
+            },
+            None => {
+                AxisAngle::new(0.0, 0.0, 0.0, 0.0)
```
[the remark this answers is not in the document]
gvaradarajan: I originally had this logic in an FFI interface function where surfacing errors is a pain, but I can have this return an error result and handle it in the FFI function again

src/spatialmath/utils.rs
gvaradarajan: I wrote it this way because it was easier for me to debug, but I can remove the helper and put the logic directly in the function

src/spatialmath/utils.rs
gvaradarajan: I don't think so, I think Rust ownership rules make both ways essentially equivalent

src/spatialmath/utils.rs
```diff
@@ -204,7 +252,7 @@ mod tests {
         let expected_ov = OrientationVector::new(
             0.0, -1.0, 0.0, 1.5707963267948966
         );
-        let calc_ov = OrientationVector::from_quaternion(&quat);
+        let calc_ov: OrientationVector = quat.into();
```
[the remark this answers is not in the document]
gvaradarajan: don't these two tests essentially provide the equivalent of a round-trip test? (the expected quaternion in one test is one of the input quaternions in the other and for the same case the opposite is true)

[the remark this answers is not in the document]
randhid: Can you show me where in person? Approving to unblock you, but I'm not sure I'm seeing that round trip logic.

randhid (approved):
LGTM, one little comment for my own knowledge to be addressed in-person.

PR #29, opened 2023-02-23
RSDK-1871 - Mdns support in rust-utils

stuqdog:
Adds mdns support to rust-utils.

Note: this won't be merged until Monday afternoon at the earliest, after a final meeting to discuss any remaining necessary tests. I figured there was no harm in getting it in front of people's eyes for review, but it may change based on the Monday meeting so folks shouldn't feel any particular pressure to review before then.


src/rpc/dial.rs
```diff
@@ -443,6 +537,54 @@ impl DialBuilder<WithCredentials> {
...
+    pub async fn connect(self) -> Result<AddAuthorization<ViamChannel>> {
+        let original_uri = match self.duplicate_uri() {
+            Some(uri) => uri,
+            None => {
+                return Err(anyhow::anyhow!(
+                    "Attempting to connect but there was no uri"
+                ))
+            }
+        };
+        let mdns_uri = webrtc::action_with_timeout(self.get_mdns_uri(), Duration::from_secs(5))
```
edaniels, line 582: I think we should reflect the same dial options go has to possible disable this. also matching its logging style feels needed to know how we are connecting

edaniels, line 582: what happens if this fails?

edaniels (changes requested):

src/rpc/dial.rs
stuqdog: `mdns_uri` becomes a `None` and we try to connect without it, using the robot uri.

stuqdog: we have `disable_mdns` in the options. if a user sets it, then `get_mdns_uri()` immediately returns a `None` and we don't try to use it. is this what you mean?

will add log statements re: how we are connecting.

edaniels: ah okay! I'm a bad reader!

stuqdog: added some log statements!

src/rpc/dial.rs
```diff
@@ -365,39 +379,132 @@ async fn get_auth_token(
...
+            return None;
+        }
+
+        log::debug!("Attempting to find address via mDNS");
+
+        let mut uri = self.duplicate_uri()?;
+        let candidate = uri.authority.clone()?.to_string();
+        let candidates = vec![candidate.replace(".", "-"), candidate.to_string()];
+
+        let stream = match mdns::discover::all(SERVICE_NAME, Duration::from_secs(1)) {
```
npmenard, line 406: would using mdns::resolve::one make more sense here? we could just target a specific robot that way

stuqdog: I tried it briefly, but found that the query was successful when using `discover::all` but unsuccessful when using `resolve::one`. I don't entirely know why this is, but suspect it's because `resolve::one` only allows a single query, but the first query doesn't seem to guarantee that we will find what we're looking for? I could be wrong, but enforcing only a single query seems to be the only difference between `resolve::one` and `discover::all` ( `resolve::one` is just doing a `discover::all` and then matching on `hostname` under the hood).

npmenard: i'm not surprised :) we should leave it this way then!

src/rpc/dial.rs
```diff
@@ -365,39 +379,132 @@ async fn get_auth_token(
...
+        let allow_downgrade = self.config.allow_downgrade;
+        if attempting_mdns {
+            log::debug!("Attempting to connect via mDNS");
+        } else {
+            log::debug!("Attempting to connect");
+        }
+        let channel = match mdns_uri {
+            Some(uri) => Self::create_channel(allow_downgrade, &domain, uri, true)
+                .await
+                .ok(),
```
npmenard, line 495: I think we should keep the error here and report it on line 505, might be helpful for debugging mdns connection failure.

npmenard (approved):
LGTM with just a comment on error reporting if direct mdns fails!

src/rpc/dial.rs
stuqdog: ahh good catch! updated

stuqdog:
Changing this back to draft so we can figure out the mDNS crate issues. 

@maximpertsov I am about to push a new commit with my (not ready for primetime) tests with the webrtc-rs crate version of mDNS.

src/rpc/dial.rs
```diff
@@ -365,39 +388,189 @@ async fn get_auth_token(
...
+                    .authority(parts.authority.clone()?)
+                    .path_and_query(parts.path_and_query.clone()?)
+                    .scheme(parts.scheme.clone()?);
+                Some(uri.build().ok()?.into_parts())
+            }
+        }
+    }
+
+    async fn get_mdns_uri(&self) -> Option<Parts> {
+        // CR erodkin: 10.1.12.11:36103
```
stuqdog, line 405: @maximpertsov this is the host and port I was expecting to get when connecting to Khari's bot.

stuqdog:
@edaniels @shawnbmccarthy this now works for both same host and not. PR to update the mdns crate, for reference: https://github.com/dylanmckay/mdns/pull/33

src/rpc/dial.rs
```diff
@@ -370,39 +388,172 @@ async fn get_auth_token(
...
+
+        let mut resp: Option<Response> = None;
+        for ipv4 in addresses {
+            let mut addr_to_send = "".to_string();
+            addr_to_send.push_str(candidates.get(1).unwrap());
+            addr_to_send.push('.');
+            addr_to_send.push_str(SERVICE_NAME);
+
+            let discovery =
+                discover::interface_with_loopback(addr_to_send, Duration::from_secs(1), ipv4)
```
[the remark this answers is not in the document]
edaniels: This is good

PR #32, opened 2023-03-13
RSDK-2268 - Update webrtc version

stuqdog:
Updates us to webrtc v0.7.1, creates a blocking `close` fn for WebRTCBaseChannel to facilitate the upgrade.

### DIFF
--- a/src/rpc/client_channel.rs
+++ b/src/rpc/client_channel.rs
@@ -21,6 +21,9 @@ use webrtc::{
 
 // see golang/client_stream.go
 const MAX_REQUEST_MESSAGE_PACKET_DATA_SIZE: usize = 16373;
+// 256 is an arbitrarily high number for maximum concurrent streams, determined based on
+// analogous value in goutils
+const MAX_CONCURRENT_STREAM_COUNT: usize = 256;
 
 /// The client-side implementation of a webRTC connection channel.
 pub struct WebRTCClientChannel {
@@ -103,11 +106,9 @@ impl WebRTCClientChannel {
     }
 
     pub(crate) fn new_stream(&self) -> Result<Stream> {
-        // 256 is an arbitrarily high number for maximum concurrent streams, determined based on
-        // analogous value in goutils
-        if self.streams.len() >= 256 {
+        if self.streams.len() >= MAX_CONCURRENT_STREAM_COUNT {
             return Err(anyhow::anyhow!(
-                "Reached max concurrent stream cap of 256; unable to add new stream."
+                "Reached max concurrent stream cap of {MAX_CONCURRENT_STREAM_COUNT}; unable to add new stream."
             ));
         }
         let id = self.stream_id_counter.fetch_add(1, Ordering::AcqRel);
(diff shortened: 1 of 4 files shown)


src/rpc/dial.rs
```diff
@@ -72,6 +72,60 @@ impl RPCCredentials {
     }
 }
 
+impl ViamChannel {
+    async fn create_resp(
```
stuqdog, line 76: note to reviewers: this is big chunk of code but it's basically just factoring out some code into a helper function. Shouldn't need too much look.

src/rpc/client_channel.rs
```diff
@@ -103,7 +102,14 @@ impl WebRTCClientChannel {
         ret_channel
     }
 
-    pub(crate) fn new_stream(&self) -> Stream {
+    pub(crate) fn new_stream(&self) -> Result<Stream> {
+        // 256 is an arbitrarily high number for maximum concurrent streams, determined based on
+        // analogous value in goutils
+        if self.streams.len() >= 256 {
```
benjirewis, line 108: [nit] parameterize 256.

src/rpc/dial.rs
```diff
@@ -88,60 +142,27 @@ impl Service<http::Request<BoxBody>> for ViamChannel {
...
-                    let response = if status_code != STATUS_CODE_OK {
-                        response.header("grpc-status", &status_code.to_string())
-                    } else {
-                        response
-                    };
+                    match channel.new_stream() {
+                        Err(e) => {
+                            log::error!("{e}");
+                            let response = response
+                                .header("grpc-status", &STATUS_CODE_UNKNOWN.to_string())
```
benjirewis, line 156: [opt] a more informative [error code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) would be RESOURCE_EXHAUSTED. Returning that here would assume that `new_stream` never returns any other kind of error, though.

src/rpc/dial.rs
```diff
@@ -88,60 +142,27 @@ impl Service<http::Request<BoxBody>> for ViamChannel {
...
                         .version(Version::HTTP_2);
 
-                    let response = if status_code != STATUS_CODE_OK {
-                        response.header("grpc-status", &status_code.to_string())
-                    } else {
-                        response
-                    };
+                    match channel.new_stream() {
+                        Err(e) => {
+                            log::error!("{e}");
```
benjirewis, line 154:
```suggestion
                            log::error!("error getting new stream: {e}");
```

benjirewis (commented):
Fixes the integer overflow on my machine. Just a couple comments.

src/rpc/dial.rs
stuqdog: ooh yeah thanks for the reminder on that one! fixed :)

src/rpc/dial.rs
stuqdog: I opted for just printing the error here because I think the error message from `new_stream()` is already pretty descriptive, so adding the prefix there just makes it feel wordy. Happy to chat more if you disagree :)

benjirewis: No need! Makes sense to me.

benjirewis (approved):
Thanks 🧑‍🔧 !

npmenard (approved):
LGTM

PR #35, opened 2023-03-21
prevent panic in `close_sync`

stuqdog:
In at least one known case (attempting to connect to a robot that is offline), `close_sync` panics when attempting to create a new thread. This PR causes to catch the panic and return it as an error instead.


njooma (approved):
srtm

npmenard (approved):
LGTM

PR #36, opened 2023-03-22
RSDK-2413 - add random reviewer workflow to rust-utils

stuqdog:


stuqdog:
@viamrobotics/sdk-netcode the `set-reviewers` job is expected to fail right now (because it's trying to grab the secret from my account) but should succeed going forward once merged into main.

PR #39, opened 2023-03-29
Rust dial test

maximpertsov:
Convert the echo example into an integration test that verifies making various echo requests (unary, streaming, bidi) over webrtc

TODO:
* [x] ~Add server startup timeout: https://github.com/viamrobotics/rust-utils/pull/39#discussion_r1153818516~
* [x] Finish porting Bi-Di test
* [x] Break each case into a separate test


tests/server/main.go
maximpertsov, line 1: Ripped from https://github.com/viamrobotics/goutils/blob/3b5719713353d0f3d2c749fad3cb9a2a30aeba04/rpc/examples/echo/server/cmd/main.go

We can probably remove many of the options from this server - long-term it might be helpful to standardize this rpc server as testing artifact (e.g. as a docker image or something).

etc/test.sh
maximpertsov, line 1: Long-term I'd like to dockerize client and server and run in a compose workflow, but for now I think a little script will do.

Makefile
```diff
@@ -1,10 +1,17 @@
...
 	cargo build
 build-example:
 	cd examples/ && cargo build
 buf-clean:
 	find src/gen -type f \( -iname "*.rs" ! -iname "mod.rs" \) -delete
-buf:	buf-clean
-	buf generate buf.build/viamrobotics/goutils --template buf.gen.yaml
-	buf generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api
+buf-install:
+	./etc/install_buf.sh $(BUF_BIN)
```
maximpertsov, line 12: See: https://github.com/viamrobotics/rust-utils/pull/39/files#r1153480965

etc/test.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+pushd tests/server && ./entrypoint.sh "$SERVER_PORT" &
+popd
+
+while ! lsof -i ":$SERVER_PORT" | grep -q LISTEN; do
+  sleep 0.1
+done
+
+SERVER_PORT=$SERVER_PORT cargo test --test "*"
```
maximpertsov, line 17: `--test "*"` makes it so we only run integration tests. without it we also run some protobuf-related tests that currently fail

etc/test.sh
maximpertsov: fair - i guess i'm thinking of a way to have generic connectivity tests that simulate different networks between client and server... but for now I just want ANY webrtc dial test that's somewhat portable. I'm going to add something similar for our js rpc in goutils.

tests/server/main.go
```diff
@@ -0,0 +1,317 @@
...
+	} else {
+		serveErr = httpServer.Serve(listener)
+	}
+	if serveErr != nil && !errors.Is(serveErr, http.ErrServerClosed) {
+		return serveErr
+	}
+	return nil
+}
+
+func computeAccessToken(privKey *rsa.PrivateKey, aud, sub string, credType rpc.CredentialsType) (string, error) {
```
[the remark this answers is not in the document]
maximpertsov: > alternatively, edit goutils and just call it directly from here with the options you need.

I prefer this approach - this is a command script in gotutils though - can publicly export that?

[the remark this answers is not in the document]
edaniels: you can git clone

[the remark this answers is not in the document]
maximpertsov: ah gotcha, okay let me try to set that up. if there are issues i will just trim the server code down as you suggested initially.

tests/echo_test.rs
```diff
@@ -0,0 +1,86 @@
...
+/// update the credentials and uri as necessary.
+use anyhow::Result;
+use std::env;
+use viam_rust_utils::gen::proto::rpc::examples::echo::v1::echo_service_client::EchoServiceClient;
+use viam_rust_utils::gen::proto::rpc::examples::echo::v1::{
+    EchoBiDiRequest, EchoMultipleRequest, EchoRequest,
+};
+use viam_rust_utils::rpc::dial;
+
+async fn dial() -> Result<dial::ViamChannel, anyhow::Error> {
```
stuqdog, line 11: (nit) you don't need to specify `anyhow::Error`, `Result<T>` is fine!

stuqdog (changes requested):
Eric got the fun stuff out of the way! One small nit on top, otherwise lgtm.

tests/echo_test.rs
maximpertsov: but writing `anyhow` is just so much fun!

jk - thank you, i will update :)

tests/server/main.go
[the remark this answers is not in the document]
maximpertsov: done https://github.com/viamrobotics/rust-utils/pull/39/commits/471dc41ffdcf2c7db3fcc398d918a985ffcec18d + https://github.com/viamrobotics/rust-utils/pull/39/commits/b5fd4ea18eb4cc86bd2c9a05f69bae8bffd4dbd8

etc/test.sh
```diff
@@ -0,0 +1,21 @@
...
+if lsof -i ":$SERVER_PORT" | grep -q LISTEN; then
+    echo "port :$SERVER_PORT is in use"
+    exit 1
+fi
+
+./etc/run_echo_server.sh "$SERVER_PORT" &
+
+while ! lsof -i ":$SERVER_PORT" | grep -q LISTEN; do
+  sleep 0.1
+done
```
maximpertsov, line 14: * [ ] timeout if server fails to start?

tests/echo_test.rs
maximpertsov: oh but doesn't this just mean i will need to specify the Result type when i call this function?

etc/test.sh
maximpertsov: Eh, it's okay - we can add this complexity later if it's a problem.

tests/echo_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert_eq!(resp.message, expected.pop().unwrap().to_string())
+    }
+    assert!(expected.is_empty());
+
+    Ok(())
+}
+
+#[ignore]
+#[tokio::test]
+async fn test_dial_webrtc_bidi() -> Result<()> {
```
maximpertsov, line 161: sigh... the webrtc bidi test is identical to the dial direct bidi test, but for some reason the stream just hangs, and does not let our logic progress to incrementing receive counts. I'm tempted to disable this one for now and make it a follow-up.

tests/echo_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert_eq!(resp.message, i.to_string());
+
+        let mut count = received.write().unwrap();
+        *count += 1;
+        drop(count);
+    }
+
+    Ok(())
+}
+
```
maximpertsov, line 108: The tests below are identical to the tests above, but test a webrtc connection. we should parameterize these in the future if they remain the same.

tests/echo_test.rs
```diff
@@ -0,0 +1,86 @@
...
+    while let Some(resp) = resp.message().await? {
+        assert_eq!(resp.message, expected.pop().unwrap().to_string())
+    }
+    assert!(expected.is_empty());
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_webrtc_bidi() -> Result<()> {
```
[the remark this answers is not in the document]
maximpertsov: update the test to interleave making requests / receiving responses. unfortunately, the webrtc version of this test hangs for reasons I'm not able to grok. would you be comfortable with me disabling the test for the moment and fixing as follow-up? alternatively, I can change it back to the non-interleaved version, which works over webrtc.

tests/echo_test.rs
maximpertsov: i could also remove the interleaving logic for now - the test passes if we make all of the rests and then receive all of the responses

etc/run_echo_server.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cd tests/goutils/rpc/examples/echo || exit 1
+
+go run server/cmd/main.go -instance-name="localhost:$1" &>/dev/null
```
maximpertsov, line 5: I disabled output from the server to make the output less noisy

etc/install_buf.sh
```diff
@@ -0,0 +1,18 @@
...
+# Substitute BIN for your bin directory.
+# Substitute VERSION for the current released version.
+
+if [[ -z "$1" ]]; then
+    echo "please provide an installation direction"
+    exit 1
+fi
+
+BIN="$1"
+VERSION="1.9.0"
```
stuqdog, line 12: (q) I may be missing context here, but why are we on `1.9.0`? It looks like latest is now `1.16.0`, do we ever expect to update?

etc/run_echo_server.sh
stuqdog: Do you think it makes sense to add an option for more verbose output, if e.g. someone wanted to run this locally?

tests/echo_test.rs
stuqdog: Why don't we dedup this now? Seems fairly straightforward and makes it all a bit cleaner.

tests/echo_test.rs
stuqdog: My vote is remove the interleaving for now if it can be done cleanly. `Some` testing is better than `None` testing. We should probably add a comment in that case though explaining the limitations of the test.

stuqdog (changes requested):

edaniels (approved):

tests/echo_test.rs
maximpertsov: > Why don't we dedup this now?

I actually had a bit of trouble parametrizing the tests, but probably due to lack of familiarity with rust.

that said, i think it might be worth keeping them separate since we the tests might diverge slightly due to legitimate dial direct / via webrtc differences. we are already going that now route by having the dial direct test interleave and having webrtc fire all and receive all: https://github.com/viamrobotics/rust-utils/pull/39/files/4cd498b6b20f45373b5634901278f97e1be49887#r1155917305

etc/run_echo_server.sh
maximpertsov: Just gonna keep it verbose for now for sake of keeping the bash scripts as light as possible: https://github.com/viamrobotics/rust-utils/pull/39/commits/dcf1b587778a0d4f9a4e21d1f029e105fe44082e

maximpertsov:
Builds failing due to upstream issue: https://github.com/viamrobotics/rust-utils/pull/40#issuecomment-1494280949

stuqdog (approved):
lgtm!

PR #40, opened 2023-03-31
RSDK-2407 - Fix gRPC error logging

stuqdog:
Major changes:
- When receiving gRPC errors, log but do not return as error (the error will still be received by users from the top level `call` function, but this avoids confusing errors that imply that any gRPC call error is a rust deserialization problem)
- Remove `close_sync` - this function was failing before because of issues creating runtimes within runtimes, which caused noisy and erroneous error messages to log when shutting down
- Add logging when dropping `WebRTCBaseChannel` to show that - despite removing the `close_sync` - we are in fact successfully dropping `WebRTCBaseChannel`s when done with them.

### DIFF
--- a/src/rpc/base_channel.rs
+++ b/src/rpc/base_channel.rs
@@ -1,7 +1,6 @@
 use anyhow::Result;
 use std::{
     fmt::Debug,
-    panic::{catch_unwind, set_hook, AssertUnwindSafe},
     sync::{
         atomic::{AtomicBool, AtomicPtr, Ordering},
         Arc,
@@ -24,6 +23,12 @@ impl Debug for WebRTCBaseChannel {
             .field("Peer connection id", &self.peer_connection.get_stats_id())
             .field("Data channel id", &self.data_channel.id())
             .finish()
+    }
+}
+
+impl Drop for WebRTCBaseChannel {
+    fn drop(&mut self) {
+        log::debug!("Dropping base channel {self:?}");
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/rpc/client_stream.rs
```diff
@@ -69,14 +69,11 @@ impl WebRTCClientStream {
...
 
-        self.base_stream.close_with_recv_error(&mut err.as_ref());
-        match err {
-            None => Ok(()),
-            Some(err) => {
-                log::error!("Error processing trailers: {err}");
-                Err(err)
-            }
+        if let Some(e) = &err {
+            log::error!("received gRPC error: {e}");
```
stuqdog, line 73: I went with `error` level logging here just to increase visibility and because it is definitely an error. However, this may be overkill since we'll still be getting an error from `call`? Happy to update to `debug` level if people think that's preferred.

edaniels: I say debug!

edaniels (approved):

src/rpc/client_stream.rs
maximpertsov: +1 for debug

src/rpc/base_channel.rs
```diff
@@ -27,6 +26,12 @@ impl Debug for WebRTCBaseChannel {
     }
 }
 
+impl Drop for WebRTCBaseChannel {
+    fn drop(&mut self) {
+        log::debug!("Dropping base channel {self:?}");
+    }
```
maximpertsov, line 32: i'm a little confused by this logic - where do we actually drop the channel object? or is the whole point of this PR that it's happening elsewhere so we should only log here?

stuqdog: `drop()` is what's called when something leaves scope. So if this calls that means the `base_channel` is definitely dropped. The log just provides verification

src/rpc/client_stream.rs
stuqdog: debug it is!

maximpertsov (approved):

npmenard (approved):
LGTM

PR #50, opened 2023-05-22
RSDK-3150 - fix ice server credential type

stuqdog:
When the webrtc crate updated to using `#[default]` instead of manually defining, a couple types had their defaults changed. One of these changes was causing stun credential failures. This updates so we no longer rely on the default.

### DIFF
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -6,7 +6,7 @@ use crate::gen::google;
 use crate::gen::proto::rpc::webrtc::v1::{
     call_response::Stage, call_update_request::Update,
     signaling_service_client::SignalingServiceClient, CallUpdateRequest,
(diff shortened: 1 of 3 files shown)


src/rpc/dial.rs
```diff
@@ -753,9 +753,10 @@ async fn maybe_connect_via_webrtc(
         Ok(resp) => resp,
         Err(e) => {
             if e.code() == tonic::Code::Unimplemented {
-                OptionalWebRtcConfigResponse::default();
+                tonic::Response::new(OptionalWebRtcConfigResponse::default())
```
stuqdog, line 756: (flyby) we weren't correctly using a default response in the `Unimplemented` case before.

cheukt (approved):

njooma (approved):

PR #51, opened 2023-05-23
[RSDK-3115] Rust + amazonlinux2

njooma:
amazonlinux2 (the default for AWS lambdas) has `glibc == 2.26`. Our release workflow was building using `glibc == 2.28`. So, I decided to use a different set of containers with a lower glibc version.

But then, it turns out our `actions-rs` actions for installing the rust toolchain and cargo were no longer supported, and causing deprecation warnings in github actions. So I updated those too. 

But then the upgraded version didn't include `cross` -- a tool for cross compiling rust libraries. The way `cross` works is by using docker containers to compile against that architecture. So I changed the container images again, this time to what `cross` was using for each of our supported platforms.

But then those images don't come with `rust` installed -- they use the host's `rust`. Well then I had to install `rust` as well.

But then the new `install rust toolchain` action was using a version of `curl` that was too advanced for the now old images. So now I install rust from source.

SO ALL IN ALL:
we now compile using `glibc == 2.17` for linux and i believe everything works as it should.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,6 +77,8 @@ jobs:
       - name: Setup rust toolchain
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          source "$HOME/.cargo/env"
+          rustup target add ${{ matrix.target }}
       - name: Setup build directory
         run: mkdir builds
       - name: Build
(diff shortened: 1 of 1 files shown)


michaellee1019 (approved):
LGTM. We did a lot of testing on AL2 within an EC2 and Lambda, these changes properly support this OS.

njooma:
@stuqdog The issue is that the base image for all AWS machines (ec2 instances, lambdas, etc) is `amazonlinux2`. The `amazonlinux2` image has an older version of glibc. So when we build our rust-utils library against our NEWER version of glibc, AWS can't run any of our rust-utils and SDKs that rely on it. 

This PR just changes how we build our release binaries to use an older version of glibc. I've tested on both amd64 and aarch64 docker images and it all works, no regressions.

PR #55, opened 2023-06-09
Remove cargo.lock from .gitignore

zaporter-work:
`rust-utils` should include the `Cargo.lock` file so that builds are deterministic and reproducible. 

See the rust book on "Cargo.toml vs Cargo.lock" (https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html):
> If you’re building a non-end product, such as a rust library that other rust packages will depend on, put Cargo.lock in your .gitignore. If you’re building an end product, which are executable like command-line tool or an application, or a system library with crate-type of staticlib or cdylib, check Cargo.lock into git. If you’re curious about why that is, see [“Why do binaries have Cargo.lock in version control, but not libraries?” in the FAQ](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

@npmenard added for visibility. I noticed that the micro-rdk also has Cargo.lock in the gitignore and I think that should probably be changed as well


stuqdog (approved):
LGTM, nice catch!

PR #56, opened 2023-06-23
RSDK-3724 - Exclude credential authorization header from ViamChannel::WebRTC based connections

edaniels:
Tested manually by making sure direct pre authorized connections still work and pass through headers while webrtc ones do not (via app or mdns found)

### DIFF
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -63,6 +63,7 @@ type SecretType = String;
 /// or a webRTC channel.
 pub enum ViamChannel {
     Direct(Channel),
+    DirectPreAuthorized(AddAuthorization<SetRequestHeader<Channel, HeaderValue>>),
     WebRTC(Arc<WebRTCClientChannel>),
 }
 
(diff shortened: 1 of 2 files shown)


src/rpc/dial.rs
```diff
@@ -143,13 +144,15 @@ impl Service<http::Request<BoxBody>> for ViamChannel {
     fn poll_ready(&mut self, cx: &mut TaskContext<'_>) -> Poll<Result<(), Self::Error>> {
         match self {
             Self::Direct(channel) => channel.poll_ready(cx),
+            Self::DirectPreAuhtorized(channel) => channel.poll_ready(cx),
```
edaniels, line 147: I don't love this but it seemed the be the simplest approach in order to get the Result type to work better without adding an Either or having it be a `dyn tonics specific service type with associations that is really annoying to type out`

src/rpc/dial.rs
```diff
@@ -143,13 +144,15 @@ impl Service<http::Request<BoxBody>> for ViamChannel {
...
             Self::WebRTC(_channel) => Poll::Ready(Ok(())),
         }
     }
 
     fn call(&mut self, request: http::Request<BoxBody>) -> Self::Future {
         match self {
             Self::Direct(channel) => Box::pin(channel.call(request)),
+            Self::DirectPreAuhtorized(channel) => Box::pin(channel.call(request)),
```
cheukt, line 155: PreAuhtorized -> PreAuthorized

stuqdog (approved):
LGTM!

src/rpc/dial.rs
edaniels: whoops

PR #57, opened 2023-06-30
dialdbg initial implementation

benjirewis:
[RSDK-3442](https://viam.atlassian.net/browse/RSDK-3442)
[RSDK-3443](https://viam.atlassian.net/browse/RSDK-3443)

cc @edaniels , @cheukt , @zaporter-work , @mattjperez 

Implements a basic CLI tool, `dialdbg`, to better understand the behavior of `rust-utils`' dial functionality (both through gRPC and WebRTC).

`dialdbg --help` output:
```
dialdbg gives information on how rust-utils' dial function makes connections

Usage: viam-dialdbg [OPTIONS] --uri <URI>

Options:
  -u, --uri <URI>
          URI to dial. Must be provided
      --nogrpc
          Whether direct gRPC connection should not be examined. If not provided, gRPC connection will be examined
      --nowebrtc
          Whether WebRTC connection should not be examined. If not provided, WebRTC connection will be examined
  -o, --output <OUTPUT>
          Filepath for output of dialdbg (file will be overwritten). If not provided, dialdbg will output to STDOUT
  -c, --credential <CREDENTIAL>
          Credential payload with which to connect to the URI. If not provided, dialdbg will dial without credentials
  -t, --credential-type <CREDENTIAL_TYPE>
          Type of credential with which to connect to the URI. Can only be provided with "--credential". If "--credential" is provided but "--credential-type" is not, credential type will default to "robot-location-secret"
  -h, --help
          Print help
  -V, --version
          Print version
```

`dialdbg --uri URI --credential ROBOTSECRET` output where `URI` is a robot on the same subnet (connectable to through mDNS) and `ROBOTSECRET` is its robot location secret.
```
Debugging dial with basic gRPC...

mDNS address 10.1.5.217:35321 was used for connection
mDNS queried in 339ms
authentication successful in 56ms
gRPC connection establishment successful in 397ms

Done debugging dial with basic gRPC.

Debugging dial with WebRTC...

mDNS address 10.1.5.217:35321 was used for connection
mDNS queried in 176ms
authentication successful in 98ms
WebRTC connection establishment successful in 345ms
selected ICE candidate pair was:
	(local) udp host e6e2f09b-bb6c-4ae4-9c5a-9f1fea22bb13.local:54432 <-> (remote) udp host 10.1.5.217:35130

nominated ICE candidates:

	remote ICE candidate:
		IP address: 127.0.0.1
		port: 49866
		nominated 1.094708ms ago
		relay protocol: udp
		network type: udp4
	local ICE candidate:
		IP address: e6e2f09b-bb6c-4ae4-9c5a-9f1fea22bb13.local
		port: 53121
		nominated 1.105ms ago
		relay protocol: udp
		network type: udp6
	remote ICE candidate:
		IP address: 100.104.116.117
		port: 54473
		nominated 1.093708ms ago
		relay protocol: udp
		network type: udp4
	local ICE candidate:
		IP address: e6e2f09b-bb6c-4ae4-9c5a-9f1fea22bb13.local
		port: 54432
		nominated 1.104291ms ago
		relay protocol: udp
		network type: udp4
	remote ICE candidate:
		IP address: 10.1.5.217
		port: 35130
		nominated 1.094375ms ago
		relay protocol: udp
		network type: udp4
	local ICE candidate:
		IP address: 71.167.222.76
		port: 60829
		nominated 1.103916ms ago
		relay protocol: udp
		network type: udp4
	remote ICE candidate:
		IP address: fd7a:115c:a1e0:ab12:4843:cd96:6268:7475
		port: 56270
		nominated 1.09525ms ago
		relay protocol: udp
		network type: udp6

Done debugging dial with WebRTC.
```

`dialdbg --uri BAD_URI --secret ROBOTSECRET` output where `bad_rover.json` is an incorrect `viam.cloud` URI.
```
Debugging dial with basic gRPC...

mDNS could not be used to connect
authentication failed
gRPC connection establishment failed

unexpected dial connect error: status: PermissionDenied, message: "failed to authenticate: no location found", details: [], metadata: MetadataMap { headers: {"content-security-policy": "frame-ancestors 'self'; form-action 'self';", "content-type": "application/grpc", "strict-transport-security": "max-age=63072000; includeSubDomains; preload", "trailer": "Grpc-Status", "trailer": "Grpc-Message", "trailer": "Grpc-Status-Details-Bin", "vary": "Origin", "x-content-type-options": "nosniff", "x-frame-options": "SAMEORIGIN", "date": "Fri, 30 Jun 2023 20:50:04 GMT", "server": "Google Frontend", "x-cloud-trace-context": "012f431cf2557357c3024279dff20029/2554969079733192940", "traceparent": "00-012f431cf2557357c3024279dff20029-237512d7599a88ec-00", "content-length": "0"} }

Done debugging dial with basic gRPC.

Debugging dial with WebRTC...

mDNS could not be used to connect
authentication failed
WebRTC connection establishment failed

unexpected dial connect error: status: PermissionDenied, message: "failed to authenticate: no location found", details: [], metadata: MetadataMap { headers: {"content-security-policy": "frame-ancestors 'self'; form-action 'self';", "content-type": "application/grpc", "strict-transport-security": "max-age=63072000; includeSubDomains; preload", "trailer": "Grpc-Status", "trailer": "Grpc-Message", "trailer": "Grpc-Status-Details-Bin", "vary": "Origin", "x-content-type-options": "nosniff", "x-frame-options": "SAMEORIGIN", "date": "Fri, 30 Jun 2023 20:50:10 GMT", "server": "Google Frontend", "traceparent": "00-196a03bfe2f8f54a3e98cb162e75396e-bc5a856d927ccbd8-00", "x-cloud-trace-context": "196a03bfe2f8f54a3e98cb162e75396e/13572307132689992664", "content-length": "0"} }

Done debugging dial with WebRTC.
```

Similar errors are output for invalid secret and invalid non-viam URI respectively:

```
unexpected dial connect error: status: Unauthenticated, message: "invalid credentials", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```
```
unexpected dial connect error: status: Unknown, message: "transport error", details: [], metadata: MetadataMap { headers: {} }
```

[RSDK-3442]: https://viam.atlassian.net/browse/RSDK-3442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-3443]: https://viam.atlassian.net/browse/RSDK-3443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Cargo.toml
```diff
@@ -54,3 +54,10 @@ env_logger = "0.9.0"
 
 [build-dependencies]
 tonic-build = {version = "0.7.2",features = ["prost", "compression"]}
+
+[workspace]
+
+# dialdbg exists as a member of the workspace so it gets its own crate.
+members = [
+    "dialdbg",
```
edaniels, line 62: can we call this `viam-dialdbg` so that you can more easily predict the `cargo run -p` invocation?

src/rpc/base_channel.rs
```diff
@@ -49,12 +52,13 @@ impl WebRTCBaseChannel {
...
                 let transport = transport.ice_transport();
                 let candidate_pair = transport.get_selected_candidate_pair().await;
-                log::info!(
-                    "Selected candidate pair. Pair: {:?}. ID: {}. Current connection state: {}",
-                    candidate_pair,
-                    conn_state,
-                    pc.get_stats_id()
-                );
+
+                // If ICE connection state is connected, log the Selected candidate pair.
```
edaniels, line 56: I think we can do this if check first before doing any of the lets/awaits

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Whether WebRTC connection should not be examined
+    #[arg(long, action, conflicts_with("nogrpc"))]
+    nowebrtc: bool,
+
+    /// Filepath for output of dialdbg (file will be overwritten)
+    #[arg(short, long)]
+    output: Option<PathBuf>,
+
+    /// Secret with which to connect to the robot
+    #[arg(short, long, required_unless_present("config"))]
```
edaniels, line 35: [minor] we should allow this to work with no secret

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+                writeln!(out, "\t\tIP address: {}", rcs.ip)?;
+                writeln!(out, "\t\tport: {}", rcs.port)?;
+                writeln!(
+                    out,
+                    "\t\tnominated {:#?} ago",
+                    now.duration_since(rcs.timestamp)
+                )?;
+                writeln!(out, "\t\trelay protocol: {}", rcs.relay_protocol)?;
+                writeln!(out, "\t\tnetwork type: {}", rcs.network_type)?;
+            }
```
edaniels, line 153: nit
```suggestion
                        stats::StatsReportType::LocalCandidate(ref cand)
            | stats::StatsReportType::RemoteCandidate(ref cand) => {
                let remote_or_local = if let stats::StatsReportType::LocalCandidate(_) = value {
                    "remote"
                } else {
                    "local"
                };
                writeln!(out, "\t{} ICE candidate:", remote_or_local)?;
                writeln!(out, "\t\tIP address: {}", cand.ip)?;
                writeln!(out, "\t\tport: {}", cand.port)?;
                writeln!(
                    out,
                    "\t\tnominated {:#?} ago",
                    now.duration_since(cand.timestamp)
                )?;
                writeln!(out, "\t\trelay protocol: {}", cand.relay_protocol)?;
                writeln!(out, "\t\tnetwork type: {}", cand.network_type)?;
            }
```

edaniels (changes requested):
Done with a first pass. nice!

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let args = Args::parse();
+
+    let mut uri = args.uri.unwrap_or_default();
+    let mut secret = args.secret.unwrap_or_default();
+    if args.config.is_some() {
+        (uri, secret) = read_config(args.config.unwrap())?;
```
zaporter-work, line 167: I think you can use `if let Some(config) = args.config {...}`

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+            }
+        },
+        None => Box::new(io::stdout()),
+    };
+
+    let mut log_config_setter = None;
+    if !args.nogrpc {
+        writeln!(out, "\nDebugging dial with basic gRPC...\n")?;
+        // Start logger with Debug-level logging and append logs to a file in a temp directory.
+        let log_path = std::env::temp_dir().join("grpc_temp.log");
```
zaporter-work, line 185: Given your bug you showed me earlier about these files not being deleted, putting something like
```
if log_path.try_exists() == Ok(true) {
   // remove log_path
}
```
might not be a bad idea in case a previous invocation errors out before this file is deleted.
https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.try_exists

dialdbg/src/parse.rs
```diff
@@ -0,0 +1,326 @@
...
+        .collect::<Vec<&str>>();
+    if split_log.len() != 2 {
+        return Err(anyhow!("malformed selected candidate message: {log}"));
+    }
+
+    // Remove annoying ": " still left over from log.
+    Ok(split_log[1].strip_prefix(": ").unwrap().to_string())
+}
+
+pub(crate) fn parse_webrtc_logs(log_path: PathBuf, out: &mut Box<dyn io::Write>) -> Result<()> {
```
zaporter-work, line 258: [opt] This function is surprisingly named. I would have expected it to return `anyhow::Result<WebRTCResult>` given its current name. Maybe something like `write_webrtc_log_statistics`? 

Similar with other functions in this file. But probably not important

zaporter-work (commented):
Mostly optional language comments. Logic looks reasonable

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    /// URI of the robot to connect to
+    #[arg(short, long, required_unless_present("config"))]
+    uri: Option<String>,
+}
+
+fn read_config(config: PathBuf) -> Result<(String, String)> {
+    let (uri, secret): (String, String);
+    let r: Value = from_str(fs::read_to_string(config)?.as_str())?;
+    if let Some(u) = r.get("address") {
```
npmenard, line 46: Would rewrite this in a more fucntional way like 
`let uri = r
        .get("address")
        .and_then(|e| e.as_str())
        .and_then(|e| Some(e.to_string()))
        .ok_or(anyhow!(
            "an error"
        ))?;`

or : 
`
let uri = r
        .get("address")
        .ok_or(anyhow!("an error"))?
        .as_str()
        .ok_or(anyhow!(
            "another error"
        ))?
        .to_string();
`

npmenard: also anyhow is great but given the repetitions of errors would consider using thiserror for reusable errors

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+        },
+        Err(e) => {
+            log::error!("{}: {e}", strings::DIAL_ERROR_PREFIX);
+            None
+        }
+    }
+}
+
+fn output_connection_stats(stats: stats::StatsReport, out: &mut Box<dyn io::Write>) -> Result<()> {
+    // StatsReport contains 13 types of stat reports; there may be more relevant stats to print
```
npmenard, line 124: would implement Display trait here instead so we can make it easier to use elswhere

src/rpc/dial.rs
```diff
@@ -467,6 +467,7 @@ impl DialBuilder<WithoutCredentials> {
         mdns_uri: Option<Parts>,
         mut original_uri_parts: Parts,
     ) -> Result<ViamChannel> {
+        log::debug!("Dialing");
```
npmenard, line 470: I would re-import the string.rs if this is possible otherwise I am concern with people accidentally remove/changing them

npmenard: if the DIALDBG_DEVELOPMENT environment variable is accessible would wrap to log statements so they are only written under test and it's clear in the code what they are meant for

npmenard (changes requested):
Done with first pass. Tricky to use log file to pass stats around (rracing could be better here) but for a POC i think its fine. I would definitely add a test to make sure that parsing works when the rust-lib gets updated.
Your parsing fails on the first error encountered, so where it makes sense I would continue parsing even in case of errors.

dialdbg/src/parse.rs
```diff
@@ -0,0 +1,326 @@
...
+
+const DEVELOPMENT: Option<&'static str> = option_env!("DIALDBG_DEVELOPMENT");
+
+#[derive(Debug, Default)]
+struct GRPCResult {
+    // The mDNS address queried (None if mDNS was not used in connection establishment).
+    mdns_address: Option<SocketAddr>,
+    // The time taken to query mDNS (None if mDNS was not used in connection establishment or
+    // query failed).
+    mdns: Option<Duration>,
```
stuqdog, line 14: (minor) `mdns` is a bit non-descriptive as a name. Can we update to be more clear? Same applies to the `WebRTCResult` type.

dialdbg/src/parse.rs
```diff
@@ -0,0 +1,326 @@
...
+    Ok(format!("{}{}", strings::DIAL_ERROR_PREFIX, split_log[1]))
+}
+
+pub(crate) fn parse_grpc_logs(log_path: PathBuf, out: &mut Box<dyn io::Write>) -> Result<()> {
+    let mut res = GRPCResult::default();
+
+    let mut connection_establishment_start = None;
+    let mut authentication_start = None;
+    let mut mdns_query_start = None;
+    for log in fs::read_to_string(log_path)?.lines() {
```
stuqdog, line 183: (q) it doesn't look to me like this validates around duplicate messages. Is this correct? If so, are there cases where we should be doing that validation?

The same for the `parse_webrtc_logs` function below.

dialdbg/src/strings.rs
```diff
@@ -0,0 +1,21 @@
+// strings.rs contains substrings of dial log messages. These constants are admittedly the most
+// fragile piece of dialdbg's code. If dialdbg is not behaving as expected, ensure these constants
```
stuqdog, line 2: (opt) can we make this less fragile by having the actual error messages point to these const strings as well? Would definitely buy "yes, but in a different PR" as a response.

stuqdog (changes requested):
First pass, looks pretty good!

dialdbg/src/main.rs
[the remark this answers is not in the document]
benjirewis: > I think this

I assume you're referring to the directing of logs to a temp file and then reading it. I'll switch to referring to the same substrings in dial and dialdbg, so we'll enforce at compile time that we output and parse the same logs. I'll add a test eventually in [RSDK-3445](https://viam.atlassian.net/browse/RSDK-3445).

src/rpc/dial.rs
benjirewis: > I would re-import the string.rs if this is possible otherwise I am concern with people accidentally remove/changing them

Good idea; will do.

> if the DIALDBG_DEVELOPMENT environment variable is accessible would wrap to log statements so they are only written under test and it's clear in the code what they are meant for

I'm a little lost on what you're suggesting here. To only `log::debug` if `DIALDBG_DEVELOPMENT` is set?

Cargo.toml
benjirewis: Interesting. `cargo run -p` outputs:
```
Possible packages/workspace members:
    viam-dialdbg
    viam-rust-utils
```
I _could_ change the workspace member name to be `viam-dialdbg` and change the directory to be called `viam-dialdbg`, but I'm not sure how idiomatic that is. I think we have a difference between directory name and package name here. `rust-utils` is the directory/repo, but the package is called `viam-rust-utils`. Similarly, I think `dialdbg` should be the directory/workspace name, and the package should be called `viam-dialdbg`. What are your thoughts, @stuqdog?

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+use viam::rpc::dial::{self, ViamChannel};
+use webrtc::stats;
+
+/// dialdbg gives information on how to connect to a Viam robot
+#[derive(Parser, Debug)]
+#[command(author, version, about)]
+struct Args {
+    /// Filepath for remote configuration JSON containing URI and secret
+    #[arg(short, long)]
+    config: Option<PathBuf>,
```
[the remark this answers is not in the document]
benjirewis: Yeah upon second thought, I'd be in favor of removing this feature (JSON config parsing) for now unless anyone disagrees. I think it's rare that a user would have the code sample "remote" JSON configuration lying around to pass to this tool, and it'd be far more likely they know the URI of their robot and the (optional) secret. @cheukt what are your thoughts?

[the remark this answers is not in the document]
cheukt: I'd be down to remove it for now

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+use std::{fs, io, path::PathBuf};
+use tokio::time::Instant;
+use viam::rpc::dial::{self, ViamChannel};
+use webrtc::stats;
+
+/// dialdbg gives information on how to connect to a Viam robot
+#[derive(Parser, Debug)]
+#[command(author, version, about)]
+struct Args {
+    /// Filepath for remote configuration JSON containing URI and secret
```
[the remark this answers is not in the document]
benjirewis: `config` was removed as an argument.

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    /// Whether direct gRPC connection should not be examined
+    #[arg(long, action, conflicts_with("nowebrtc"))]
+    nogrpc: bool,
+
+    /// Whether WebRTC connection should not be examined
+    #[arg(long, action, conflicts_with("nogrpc"))]
+    nowebrtc: bool,
+
+    /// Filepath for output of dialdbg (file will be overwritten)
```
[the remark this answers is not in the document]
benjirewis: Elaborated on all of these docstrings for behavior when not providing.

dialdbg/src/main.rs
benjirewis: Done.

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    secret: Option<String>,
+
+    /// URI of the robot to connect to
+    #[arg(short, long, required_unless_present("config"))]
+    uri: Option<String>,
+}
+
+fn read_config(config: PathBuf) -> Result<(String, String)> {
+    let (uri, secret): (String, String);
+    let r: Value = from_str(fs::read_to_string(config)?.as_str())?;
```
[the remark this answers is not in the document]
benjirewis: This function no longer exists. Let me know if you see any other places where I used unnecessarily short variable names.

dialdbg/src/main.rs
benjirewis: This function no longer exists.

> also anyhow is great but given the repetitions of errors would consider using thiserror for reusable errors

Will look into this! Thanks.

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    } else {
+        return Err(anyhow!("no top-level 'secret' value in provided config"));
+    }
+    Ok((uri, secret))
+}
+
+async fn dial_grpc(uri: &str, secret: &str) {
+    let creds = dial::RPCCredentials::new(
+        None,
+        "robot-location-secret".to_string(),
```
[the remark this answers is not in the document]
benjirewis: Good point; added a new `secret_type` argument (specified as `--secret-type` or `-t`).

dialdbg/src/main.rs
benjirewis: I could do this, but to clarify I would need to make a new type that wraps `stats::StatsReport` and implement `Display` for that type (both the struct and trait are foreign).

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+fn output_connection_stats(stats: stats::StatsReport, out: &mut Box<dyn io::Write>) -> Result<()> {
+    // StatsReport contains 13 types of stat reports; there may be more relevant stats to print
+    // here, but for now I have stuck with only printing the candidates.
+    writeln!(out, "\nnominated ICE candidates:\n")?;
+    let now = Instant::now();
+    for value in stats.reports.into_values() {
+        match value {
+            stats::StatsReportType::LocalCandidate(lcs) => {
```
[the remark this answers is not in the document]
benjirewis: Good point; neither the debug-print nor JSON form of `ICECandidateStats` looks very nice... So I'll stick with writelns of various fields for now.

dialdbg/src/main.rs
benjirewis: Moved local and remote to one arm as described above. The `matches` macro kind of scared me haha, but if you feel strongly it's a better solution, @stuqdog, I can look more into it.

dialdbg/src/main.rs
npmenard: if you do it like `pub struct StatsReport(stats::StatsReport)`

dialdbg/src/main.rs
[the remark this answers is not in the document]
edaniels: So maybe we should also be calling this credential type along with the word payload which is what it's actually called?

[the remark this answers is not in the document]
benjirewis: I don't know... I could see arguments for both current naming and your new suggested one. Argument for being that we're consistent with underlying rust-utils and proto naming. Argument against being that users and devs are probably familiar with "secret" and therefore "secret type" from app UI and docs.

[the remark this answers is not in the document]
edaniels: well if the tool is just for app managed viam robots, then secret is fine. if it's for pure open source cases (no app), then I feel like the proposed way is better.

dialdbg/src/main.rs
edaniels: @stuqdog just for my own edification, what would be better about the macro?

dialdbg/src/main.rs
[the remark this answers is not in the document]
edaniels: cool

src/rpc/dial.rs
npmenard: > > I would re-import the string.rs if this is possible otherwise I am concern with people accidentally remove/changing them
> 
> Good idea; will do.
> 
> > if the DIALDBG_DEVELOPMENT environment variable is accessible would wrap to log statements so they are only written under test and it's clear in the code what they are meant for
> 
> I'm a little lost on what you're suggesting here. To only `log::debug` if `DIALDBG_DEVELOPMENT` is set?

I was thinking of a macro looking like this : 
`
macro_rules! dbg_log {

    ($($args: tt)*) => {
        if std::env::var_os("DIALDBG").is_some(){
            log::debug!($($args)*);
        }

    }
}
`

dialdbg/src/main.rs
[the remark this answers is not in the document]
benjirewis: I think it's hypothetically for pure open source cases. Removed all references to `robot` and `secret`. Changed `secret` to `credential` and `secret-type` to `credential-type`. `payload` feels like a fairly vague word, and `credential-payload` feels too long.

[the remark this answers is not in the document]
edaniels: hah fair

Cargo.toml
edaniels: eh i dont feel strongly!

edaniels (approved):
LGTM from my side but I am yet a lowly rust newbie.

dialdbg/src/main.rs
stuqdog: @benjirewis I don't feel strongly, this looks fine!
@edaniels it's not a big difference, but currently we have the catch-all no-op case at the bottom, which we can avoid with `if matches!(<var>, local | remote) { ... }`. Nothing major, just a bit cleaner imo!

Cargo.toml
stuqdog: For consistency's sake/to avoid modifying the existing `rust-utils` structure, I think I favor keeping it as-is, but it's not a strong preference!

src/rpc/dial.rs
benjirewis: I like the idea of using a macro, but my idea was that `DIALDBG_DEVELOPMENT` was to be used to print rust-utils' logs to dialdbg's output, so that I could see what rust-utils was outputting during development of dialdbg. What you've described above seems like it'll log rust-utils' logs only if `DIALDBG_DEVELOPMENT` is defined. I _could_ restrict rust-utils' logs (or a subset of them) to only `log::debug!` if some environment variable is set, but that feels strange. In my head, the debug messages I've added, like "Dialing", were no different from the existing, "normal" log statements.

dialdbg/src/main.rs
benjirewis: Without `read_config`, I'm not sure there's enough reused errors in `parse.rs` to warrant a stronger error system right now. Let me know if you still feel strongly.

dialdbg/src/main.rs
benjirewis: Yeah I couldn't find a way to use the `matches` macro _and_ extract the actual candidate from the `LocalCandidate` / `RemoteCandidate` enum. i.e. couldn't do `if matches!(report_value, LocalCandidate(cand) | RemoteCandidate(cand) { ...access cand })`. Maybe I'm missing something, though.

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    let mut uri = args.uri.unwrap_or_default();
+    let mut secret = args.secret.unwrap_or_default();
+    if args.config.is_some() {
+        (uri, secret) = read_config(args.config.unwrap())?;
+    }
+
+    // Write to output file or STDOUT if none is provided.
+    let mut out: Box<dyn io::Write> = match args.output {
+        Some(output) => match fs::File::create(output) {
```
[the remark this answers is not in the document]
benjirewis: From what I understand, `map` would change a `Result<T, E>` to `Result<U, E>` (allows me to mutate the underlying Ok value). While I do want to wrap the underlying Ok value in a `Box`, I _also_ want to wrap any underlying Err value with some context. I have a _feeling_ I could use `map_or_else`, but maybe I'm missing what you're meaning here?

dialdbg/src/parse.rs
benjirewis: Updated to `mdns_query`, but I don't know if that's any more descriptive. I could suffix most of these fields with `time` or `duration` if that sounds better to you?

dialdbg/src/parse.rs
benjirewis: A great point. Right now we simply honor the last message of its kind in setting fields on result types. I don't think `rust-utils` would ever output duplicates of the messages we're interested in unless something was wrong with dial. I _could_ add validation, but it's not "trivial" (makes the code look a little messy). I think right now, the last-message-honored approach is probably ok, but I'm happy to add duplicate errors if you think it's more robust.

dialdbg/src/parse.rs
benjirewis: Good point; Modified them to return `Result<*Result>` and actually printed the structs in `main.rs`.

dialdbg/src/strings.rs
benjirewis: Yep! Added. @npmenard requested the same thing. I moved the strings to `rpc/log_prefixes` so we don't have a circular dependency.

src/rpc/dial.rs
```diff
@@ -527,12 +528,14 @@ impl DialBuilder<WithoutCredentials> {
...
 
         if disable_webrtc {
+            log::debug!("Connected via gRPC");
             Ok(ViamChannel::Direct(channel.clone()))
         } else {
             match maybe_connect_via_webrtc(uri, intercepted_channel.clone(), webrtc_options).await {
                 Ok(webrtc_channel) => Ok(ViamChannel::WebRTC(webrtc_channel)),
                 Err(e) => {
                     log::error!("error connecting via webrtc: {e}. Attempting to connect directly");
+                    log::debug!("Connected via gRPC");
```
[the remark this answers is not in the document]
benjirewis: Let's discuss offline! I think I'm prematurely declaring gRPC connection success in a couple places.

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    if let Some(u) = r.get("address") {
+        // We use some funky as_str -> to_string logic as serde_json weirdly wraps to_string'ed
+        // Values in quotes. Also, we want to verify "address" and "secret" are actually strings.
+        // Was I too lazy to use regular serde? Yes.
+        if let Some(uri_str) = u.as_str() {
+            uri = uri_str.to_string();
+        } else {
+            return Err(anyhow!(
+                "expected 'address' value to be a string in provided config"
+            ));
```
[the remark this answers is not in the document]
benjirewis: Great point! Switched to bail everywhere.

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+
+fn read_config(config: PathBuf) -> Result<(String, String)> {
+    let (uri, secret): (String, String);
+    let r: Value = from_str(fs::read_to_string(config)?.as_str())?;
+    if let Some(u) = r.get("address") {
+        // We use some funky as_str -> to_string logic as serde_json weirdly wraps to_string'ed
+        // Values in quotes. Also, we want to verify "address" and "secret" are actually strings.
+        // Was I too lazy to use regular serde? Yes.
+        if let Some(uri_str) = u.as_str() {
+            uri = uri_str.to_string();
```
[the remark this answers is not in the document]
benjirewis: Function is gone, but yeah it was strange 🤔 .

dialdbg/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+                writeln!(out, "\t\tport: {}", lcs.port)?;
+                writeln!(
+                    out,
+                    "\t\tnominated {:#?} ago",
+                    now.duration_since(lcs.timestamp)
+                )?;
+                writeln!(out, "\t\trelay protocol: {}", lcs.relay_protocol)?;
+                writeln!(out, "\t\tnetwork type: {}", lcs.network_type)?;
+            }
+            stats::StatsReportType::RemoteCandidate(rcs) => {
```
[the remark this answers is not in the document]
benjirewis: JSONifying the output of `dialdbg` is a good idea, but I'll probably save it as future work for now. We're not aware of any immediate use cases for programmatically processing `dialdbg`s output.

src/rpc/dial.rs
npmenard: I see I didn't realize we would have the logs on most (all) the time make sense to not wrap them then

npmenard (approved):
LGTM

src/rpc/dial.rs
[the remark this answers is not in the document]
benjirewis: Discussed offline; this is related to [a bug](https://viam.atlassian.net/browse/RSDK-3846) @npmenard reported a while ago. rust-utils can return a gRPC connection that isn't exactly a "successfully established" connection. I don't think I'll fix that issue in this PR.

zaporter-work:
Don't have time for in-depth review, but all overall LGTM!

src/rpc/dial.rs
```diff
@@ -673,10 +683,15 @@ impl DialBuilder<WithCredentials> {
...
             }
         };
-        let mdns_uri = webrtc::action_with_timeout(self.get_mdns_uri(), Duration::from_secs(5))
-            .await
-            .ok()
-            .flatten();
+        // NOTE(benjirewis): Use a duration of 1500ms for getting the mDNS URI. I've anecdotally
+        // seen times as great as 922ms to fetch a non-loopback mDNS URI. With an
+        // interface_with_loopback query interval of 250ms, 1500ms here should give us time for ~6
+        // queries.
```
benjirewis, line 689: @cheukt / @stuqdog , I lowered both the `interface_with_loopback` query interval and the `get_mdns_uri` timeout  to allow for faster-completing mDNS discovery (@erh hopefully this fixes your issue with excessively long dials when mDNS is enabled but not used).

dialdbg/src/main.rs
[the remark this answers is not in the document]
stuqdog: I think we could do
```
fs::File::create(output)
.map(Box::new)
.map_err(|e| anyhow::anyhow!("error opening --output file: {e}")?
```
(not tested, might be slightly wrong)
`map_or_else` would also work I think but it'd be a slight code smell imo. but it's not a big deal whichever way, feel free to leave as-is if you prefer!

dialdbg/src/parse.rs
```diff
@@ -0,0 +1,335 @@
...
+use viam::rpc::log_prefixes;
+
+const DEVELOPMENT: Option<&'static str> = option_env!("DIALDBG_DEVELOPMENT");
+
+// This prefix is prepended in dialdbg when connect returns an error. It is not
+// from dial itself.
+pub(crate) const DIAL_ERROR_PREFIX: &'static str = "unexpected dial connect error";
+
+#[derive(Debug, Default)]
+pub(crate) struct GRPCResult {
```
stuqdog, line 13: Probably over-engineering right now, but it would be neat if we could dedup some of the shared code across `GRPCResult` and `WebRTCResult`

dialdbg/src/parse.rs
stuqdog: I think it's fine to leave as is, thanks for clarifying! We can always add duplicate error checking down the line.

stuqdog (approved):
LGTM!

dialdbg/src/main.rs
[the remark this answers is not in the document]
benjirewis: Ah yeah, the combo of `map` and `map_err` definitely works; thanks! `map_or_else` has that extra default field, so I agree it's a smell.

dialdbg/src/parse.rs
benjirewis: That's a good point; I probably won't in this PR, but I can imagine deduping more in the future as the code around these structs increases (potentially as part of [RSDK-3444](https://viam.atlassian.net/browse/RSDK-3444)).

PR #59, opened 2023-07-05
RSDK-3833 Dial takes type and payload as parameters

jckras:
Credential type is currently hard-coded as robot-location-secret, but we’re starting to also use robot-secret. In order to pass type as a parameter in the cpp-sdk we need to update the dial function in rust-utils to account for the new argument also.

### DIFF
--- a/examples/src/ffi/cpp/ffi_robot.cc
+++ b/examples/src/ffi/cpp/ffi_robot.cc
@@ -19,7 +19,7 @@ using viam::robot::v1::ResourceNamesResponse;
 extern "C" void *init_rust_runtime();
 extern "C" int free_rust_runtime(void *ptr);
 extern "C" void free_string(char* s);
-extern "C" char *dial(const char *uri, const char *payload,
+extern "C" char *dial(const char *uri, const char *r#type, const char *payload,
                              bool allow_insecure, void *ptr);
 
 class RobotServiceClient {
@@ -51,7 +51,7 @@ private:
 int main(int argc, char *argv[]) {
 
   void *ptr = init_rust_runtime();
-  char *path = dial("<your robot's uri>", "<your robot's credentials>", false, ptr);
+  char *path = dial("<your robot's uri>", "<your robot's payload>", "<your robot's type>", false, ptr);
   if(path == NULL){
 	  free_rust_runtime(ptr);
 	  return 1;
--- a/src/ffi/dial_ffi.rs
+++ b/src/ffi/dial_ffi.rs
@@ -10,6 +10,7 @@ use tokio::runtime::Runtime;
 use tokio::sync::oneshot;
 use tracing::Level;
 
+
 use crate::rpc::dial::{
     DialBuilder, DialOptions, RPCCredentials, ViamChannel, WithCredentials, WithoutCredentials,
 };
@@ -87,13 +88,14 @@ fn dial_without_cred(
 
 fn dial_with_cred(
     uri: String,
+    r#type: &str,
     payload: &str,
     allow_insec: bool,
     disable_webrtc: bool,
 ) -> Result<DialBuilder<WithCredentials>> {
     let creds = RPCCredentials::new(
         None,
-        String::from("robot-location-secret"),
+        String::from(r#type),
         String::from(payload),
     );
     let c = DialOptions::builder().uri(&uri).with_credentials(creds);
@@ -113,12 +115,14 @@ fn dial_with_cred(
 /// When falling to dial it will return a NULL pointer
 /// # Arguments
 /// * `c_uri` a C-style string representing the address of robot you want to connect to
+/// * `c_type` a C-style string representing the type of robot's secret you want to use, set to NULL if you don't need authentication
 /// * `c_payload` a C-style string that is the robot's secret, set to NULL if you don't need authentication
 /// * `c_allow_insecure` a bool, set to true when allowing insecure connection to your robot
 /// * `rt_ptr` a pointer to a rust runtime previously obtained with init_rust_runtime
 #[no_mangle]
 pub unsafe extern "C" fn dial(
     c_uri: *const c_char,
+    c_type: *const c_char,
     c_payload: *const c_char,
     c_allow_insec: bool,
     rt_ptr: Option<&mut DialFfi>,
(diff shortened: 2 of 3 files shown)


examples/src/ffi/cpp/ffi_robot.cc
```diff
@@ -51,7 +51,7 @@ class RobotServiceClient {
 int main(int argc, char *argv[]) {
 
   void *ptr = init_rust_runtime();
-  char *path = dial("<your robot's uri>", "<your robot's credentials>", false, ptr);
+  char *path = dial("<your robot's uri>", "<your robot's payload>", "<your robot's type>", false, ptr);
```
cheukt, line 54: isn't this switched from above given the signature is `type` then `payload`?

src/ffi/dial_ffi.rs
```diff
@@ -137,6 +141,21 @@ pub unsafe extern "C" fn dial(
         ur
     };
     let allow_insec = c_allow_insec;
+    if!c_payload.is_null() && c_type.is_null(){
+        println!("Error missing credential: secret type");
```
stuqdog, line 145: I don't think we should be using `println` in production code. Can we do error logging instead?

Probably would be good to replace any existing `println` lines in this file with `log` statements as well, while we're at it.

src/ffi/dial_ffi.rs
```diff
@@ -137,6 +141,21 @@ pub unsafe extern "C" fn dial(
...
+        return ptr::null_mut();
+
+    }
+    if c_payload.is_null() && !c_type.is_null(){
```
stuqdog, line 149: These checks feel a bit redundant given what's happening below at line 202. And, though we define `r#type` and `payload` here, they don't actually get used until line 202. Can we just combine all this logic at a single point so we're not doing this "is it null?" check and also a "is it None?" check?

stuqdog (changes requested):

examples/src/ffi/cpp/ffi_robot.cc
jckras: Good catch thank you!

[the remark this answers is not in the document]
jckras: What is the difference between a robot type and a authentication type? Wouldn't a user always need to be authenticated to connect to a robot? (In other words, how would someone have a type and payload to enter but no robot?)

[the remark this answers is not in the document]
stuqdog: Well, let's say we have 3 authentication types: robot secret, location secret, and user secret. All of them will let you authenticate and access a robot, but the location and user secrets are only tangentially related to the robot (the robot is in the location, but so are lots of robots). There is no 1:1 saying "_this_ secret means specifically authenticating to _that_ robot".

Also, when i see "robot type", I think, "is it a farmer robot? a delivery robot? a vacuuming robot?". What we want here isn't the _robot_ type, it's the _authentication_ type (which again, could be based on a specific robot, or a location, or hypothetically other things down the line).

src/ffi/dial_ffi.rs
```diff
@@ -27,6 +27,7 @@ use tower_http::{
 
 use anyhow::Result;
 
+
```
stuqdog, line 30: (nit) probably should get rid of this newline.

src/ffi/dial_ffi.rs
```diff
@@ -205,10 +197,17 @@ pub unsafe extern "C" fn dial(
...
                     .await?
             }
-            _ => {
+            (None, None) => {
                 let c = dial_without_cred(uri_str, allow_insec, disable_webrtc)?;
                 c.connect().await?
-                
+            }
+            (None, Some(_)) => {
+                // the return type here is finicky, so we are taking advantage of the ? operator to return an errenous value
```
stuqdog, line 205: I know I suggested a comment like this, but on reflection I think it's a bit awkward and unclear. I think a better pattern would be to not use the `?` operator inside the `match` statement, but rather immediately outside of it.

stuqdog (changes requested):

src/ffi/dial_ffi.rs
jckras: Sorry I missed this, thank you for the reminder! It will be updated in the next commit.

examples/src/ffi/cpp/ffi_robot.cc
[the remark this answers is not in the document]
jckras: Ah got it. So a location secret could be an authentication to the robot or someone might only have access to a robot secret and no location at all. And to your second point, people might actually have different types of robots - that would be really confusing. I am going to update in cpp and python also :)

cheukt (approved):
lgtm if manually tested

stuqdog (changes requested):
A couple small nits, otherwise looks great! 

Agree with Cheuk that I'll want to know that we've tested the latest and greatest version of this manually, just to make sure we didn't introduce any regressions. Also, we should definitely coordinate release here to make sure C++ and python are updated at the same time.

jckras:
> A couple small nits, otherwise looks great!
> 
> Agree with Cheuk that I'll want to know that we've tested the latest and greatest version of this manually, just to make sure we didn't introduce any regressions. Also, we should definitely coordinate release here to make sure C++ and python are updated at the same time.

I was testing this by copying the dylib and running the cpp/python dial example, is that the correct way to manually test or should I be running an a example from within rust-utils code itself?

stuqdog:
> > A couple small nits, otherwise looks great!
> > Agree with Cheuk that I'll want to know that we've tested the latest and greatest version of this manually, just to make sure we didn't introduce any regressions. Also, we should definitely coordinate release here to make sure C++ and python are updated at the same time.
> 
> I was testing this by copying the dylib and running the cpp/python dial example, is that the correct way to manually test or should I be running an a example from within rust-utils code itself?

Yeah that works! Just want to make sure our testing includes all the latest changes.

benjirewis (approved):
LGTM % outstanding nits. Great job taking on some Rust @jckras 🦀 !

jckras:
> > > A couple small nits, otherwise looks great!
> > > Agree with Cheuk that I'll want to know that we've tested the latest and greatest version of this manually, just to make sure we didn't introduce any regressions. Also, we should definitely coordinate release here to make sure C++ and python are updated at the same time.
> > 
> > 
> > I was testing this by copying the dylib and running the cpp/python dial example, is that the correct way to manually test or should I be running an a example from within rust-utils code itself?
> 
> Yeah that works! Just want to make sure our testing includes all the latest changes.

Waiting to test on python and cpp-sdk and then will re-commit

jckras:
> > > A couple small nits, otherwise looks great!
> > > Agree with Cheuk that I'll want to know that we've tested the latest and greatest version of this manually, just to make sure we didn't introduce any regressions. Also, we should definitely coordinate release here to make sure C++ and python are updated at the same time.
> > 
> > 
> > I was testing this by copying the dylib and running the cpp/python dial example, is that the correct way to manually test or should I be running an a example from within rust-utils code itself?
> 
> Yeah that works! Just want to make sure our testing includes all the latest changes.

Just tested rust-utils manually in python and cpp and in both cases we are connecting to the viam server

stuqdog (approved):
Looks great, thanks for all your work on this!

PR #61, opened 2023-07-13
RSDK-4011 Add more mDNS information to dialdbg

benjirewis:
[RSDK-4011](https://viam.atlassian.net/browse/RSDK-4011).

If mDNS was not used to connect; lists all mDNS responses returned for the current subnet. Also queries candidates with `-`  instead of `.`.

[RSDK-4011]: https://viam.atlassian.net/browse/RSDK-4011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dialdbg/src/main.rs
+++ b/dialdbg/src/main.rs
@@ -126,14 +126,13 @@ async fn dial_webrtc(
 const MDNS_SERVICE_NAME: &'static str = "_rpc._tcp.local";
 
 async fn output_all_mdns_addresses(out: &mut Box<dyn io::Write>) -> Result<()> {
-    writeln!(out, "\ndiscovered mDNS addresses:")?;
-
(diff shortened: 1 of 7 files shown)


dialdbg/src/main.rs
```diff
@@ -121,6 +122,48 @@ async fn dial_webrtc(
     }
 }
 
+const MDNS_SERVICE_NAME: &'static str = "_rpc._tcp.local";
```
stuqdog, line 125: since we're already using values from `viam::rpc::dial`, can we just use the `SERVICE_NAME` const from there instead of recreating it here?

stuqdog (approved):

PR #62, opened 2023-07-14
RSDK-4034 - Fix cargo bump

stuqdog:
#### Major Changes
Replace use of `cargo-bump` (which does not play well with workspaces) with `cargo-release` (which does)

#### Minor Changes
Commit example `Cargo.lock`, which was missed when we removed `cargo.lock` from the `.gitignore`


Cargo.toml
```diff
@@ -61,3 +61,4 @@ tonic-build = {version = "0.9.2",features = ["prost"]}
 members = [
     "dialdbg",
 ]
+exclude = [ "examples" ]
```
stuqdog, line 64: (flyby) examples thought of themselves as part of the workspace, need to exclude them or else they won't run.

cheukt (approved):

PR #65, opened 2023-07-19
RDSK-3444 Add basic RTT measuring to dialdbg

benjirewis:
[RSDK-3444](https://viam.atlassian.net/browse/RSDK-3444)

Adds basic RTT measuring to `dialdbg` (can be disabled with `--nortt`. When a connection is successfully established; 10 echo requests will be run using the created channel, and the average RTT will be printed as:

```
average RTT across established [type of connection] connection: 149ms
```

[RSDK-3444]: https://viam.atlassian.net/browse/RSDK-3444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dialdbg/src/main.rs
+++ b/dialdbg/src/main.rs
@@ -205,7 +205,7 @@ async fn main() -> Result<()> {
 
         if let Some(ch) = ch {
             if !args.nortt {
-                let average_rtt = rtt::measure_rtt(ch.clone(), 10).await?;
+                let average_rtt = rtt::measure_rtt(ch, 10).await?;
                 writeln!(
                     out,
                     "average RTT across established gRPC connection: {}ms",
--- a/dialdbg/src/main.rs
+++ b/dialdbg/src/main.rs
@@ -1,4 +1,5 @@
 mod parse;
+mod rtt;
 mod stats;
 
 use anyhow::{anyhow, Result};
@@ -23,6 +24,11 @@ struct Args {
     #[arg(long, action, conflicts_with("nogrpc"))]
     nowebrtc: bool,
 
+    /// Whether round-trip-time across established connections should be measured. If not provided,
+    /// round-time-time will be measured.
+    #[arg(long, action)]
+    nortt: bool,
+
     /// Filepath for output of dialdbg (file will be overwritten). If not provided, dialdbg will
     /// output to STDOUT.
     #[arg(short, long)]
@@ -44,7 +50,7 @@ struct Args {
     uri: Option<String>,
 }
 
-async fn dial_grpc(uri: &str, credential: &str, credential_type: &str) {
+async fn dial_grpc(uri: &str, credential: &str, credential_type: &str) -> Option<ViamChannel> {
     let dial_result = match credential {
         "" => {
             dial::DialOptions::builder()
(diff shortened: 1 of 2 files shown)


dialdbg/src/rtt.rs
```diff
@@ -0,0 +1,28 @@
...
+use viam::rpc::dial::ViamChannel;
+
+// Returns the average round-trip-time over num_pings for the passed-in channel.
+pub(crate) async fn measure_rtt(ch: ViamChannel, num_pings: u32) -> Result<time::Duration> {
+    let mut total_ping = time::Duration::new(0, 0);
+    for _ in 0..num_pings {
+        let start = time::Instant::now();
+
+        // Send an echo request across the channel. It's unlikely the remote will be able to
+        // respond to this request, but we'll still get a good sense of RTT.
```
benjirewis, line 15: I don't feel great about this strategy; I wish there were a request we could send across the channel that would never result in an error (notice the `ok()` call below to get rid of `Unimplemented` errors). In practice, this does seem to measure an RTT. To a rover in the office: ~120ms via gRPC and ~140ms via WebRTC. To a server on my machine: ~2ms via gRPC and ~1ms via WebRTC. Do let me know if you have suggestions for improvements here.

stuqdog: I actually kind of think `echo` is good here. Ideally we want a no-op so we're only accounting for the actual network time and not for server-side processing time, but I don't think we can do that right now in rust. An unimplemented `echo` request seems like a decent enough proxy.

dialdbg/src/main.rs
```diff
@@ -197,10 +199,21 @@ async fn main() -> Result<()> {
             )?;
         log_config_setter = Some(log4rs::init_config(config)?);
 
-        dial_grpc(uri.as_str(), credential.as_str(), credential_type.as_str()).await;
+        let ch = dial_grpc(uri.as_str(), credential.as_str(), credential_type.as_str()).await;
```
stuqdog, line 202: (nit) preference for moving this down to line 206 where it's being used (and optionally just combining the two lines `if let Some(ch) = dial_grpc(...)`).

dialdbg/src/main.rs
```diff
@@ -197,10 +199,21 @@ async fn main() -> Result<()> {
...
         let grpc_res = parse::parse_grpc_logs(log_path.clone(), &mut out)?;
         write!(out, "{grpc_res}")?;
 
+        if let Some(ch) = ch {
+            if !args.nortt {
+                let average_rtt = rtt::measure_rtt(ch.clone(), 10).await?;
```
stuqdog, line 208: (nit) I don't think this call to `clone()` is necessary?

stuqdog (approved):
lgtm % a couple small nits!

dialdbg/src/main.rs
benjirewis: Good point; normally I'd agree, but I just wanted the parsed log information to appear before average RTT.

dialdbg/src/main.rs
benjirewis: removed.

dialdbg/src/main.rs
```diff
@@ -235,20 +248,32 @@ async fn main() -> Result<()> {
             log4rs::init_config(config)?;
         }
 
-        let sr = dial_webrtc(uri.as_str(), credential.as_str(), credential_type.as_str()).await;
+        let ch = dial_webrtc(uri.as_str(), credential.as_str(), credential_type.as_str()).await;
```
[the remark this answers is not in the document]
benjirewis: See above comment; I was trying to get a certain ordering of the output.

npmenard (approved):
LGTM

dialdbg/src/main.rs
stuqdog: Ahh sure, that makes sense. Okay, disregard!

maximpertsov:
Pardon the late review - looks good!

PR #67, opened 2023-07-25
RSDK-3446 Add build-dialdbg makefile target

benjirewis:
RSDK-3446

Adds `build-dialdbg` makefile target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ build:
 	cargo build
 build-example:
 	cd examples/ && cargo build
+build-dialdbg:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,6 +6,8 @@ build:
 	cargo build
 build-example:
 	cd examples/ && cargo build
+build-dialdbg:
```
benjirewis, line 9: I could also add this target to the `all` target but felt it wasn't "part" of `rust-utils` in the same way as the examples.

PR #73, opened 2023-08-24
RSDK-1996 - use install_name_tool to fix the install path

stuqdog:
#### Major changes
Use `install_name_tool` in build action to override the github artifact install name path.

#### Testing
Ran the `install_name_tool` command on a dylib downloaded from github, observed corrected behavior when working with the C++ SDK.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,7 +109,7 @@ jobs:
         run: cp target/${{ matrix.target }}/release/libviam_rust_utils.dylib builds/libviam_rust_utils-${{ matrix.platform }}.dylib
       - name: Correct install path
         run: |
-          install_name_tool -id "@rpath/libviam_rust_utils-${{ matrix.platform }}.dylib" builds/libviam_rust_utils-${{ matrix.platform }}.dylib
+          install_name_tool -id "@rpath/libviam_rust_utils.dylib" builds/libviam_rust_utils-${{ matrix.platform }}.dylib
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


stuqdog:
@andf-viam fyi

.github/workflows/release.yml
```diff
@@ -107,6 +107,9 @@ jobs:
           cargo build --release --target=${{ matrix.target }}
       - name: Copy
         run: cp target/${{ matrix.target }}/release/libviam_rust_utils.dylib builds/libviam_rust_utils-${{ matrix.platform }}.dylib
+      - name: Correct install path
+        run: |
+          install_name_tool -id "@rpath/libviam_rust_utils-${{ matrix.platform }}.dylib" builds/libviam_rust_utils-${{ matrix.platform }}.dylib
```
acmorrow, line 112: Two things here:

- Setting the install name to `@rpath/libviam_rust_utils-${{ matrix.platform }}.dylib` means that the file must be on disk with the `-{{matrix.platform}}` component on the name. That isn't the current practice in the C++ SDK, where the file is installed as `libviam_rust_utils.dylib`.

- I don't think `@rpath` is entirely sufficient. There is nothing in the C++ SDK build that will put `$PREFIX/lib` or `../lib` into the rpath stack, which would be required for this to resolve. Now, maybe the C++ SDK could (should?) do that in order to make it work.

stuqdog: Good catch, thanks! Fixed the filename issue. As to whether `@rpath` is sufficient, we can work on fixing that in the SDK?

acmorrow (approved):
Let's move forward with this, and figure out what needs to be done C++ SDK side. Worst case we revisit this, but in more or less all cases having this file switch to an install name with `@rpath` is an improvement over the status quo with the hardcoded github actions path.

LGTM.

PR #80, opened 2023-09-25
RSDK-4447: Add dial timeout

hexbabe:
[RSDK-4447](https://viam.atlassian.net/browse/RSDK-4447?atlOrigin=eyJpIjoiYmMxYjQwMGQ3ZTAwNGEzMDhjNjA4OTMzOTRmNzgwZjciLCJwIjoiaiJ9)

[Related changes in Python SDK](https://github.com/viamrobotics/viam-python-sdk/pull/444)

[RSDK-4447]: https://viam.atlassian.net/browse/RSDK-4447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Test procedure

- [Test with client that connects to offline robot](https://www.loom.com/share/a8a5661dd1304c55bdbd4ae4e628368c)
- [Test with robot that is online, but edited the rust-utils code to sleep at the end of the answerer code](https://www.loom.com/share/b0f459a1bc674a8b976468ac976f4781?sid=381570a9-fa45-4917-984d-ef1523073312)

### DIFF
--- a/src/ffi/dial_ffi.rs
+++ b/src/ffi/dial_ffi.rs
@@ -116,7 +116,7 @@ fn dial_with_cred(
 /// * `c_type` a C-style string representing the type of robot's secret you want to use, set to NULL if you don't need authentication
 /// * `c_payload` a C-style string that is the robot's secret, set to NULL if you don't need authentication
 /// * `c_allow_insecure` a bool, set to true when allowing insecure connection to your robot
-/// * `c_timeout` a float, set how long we should try dial before timing out
+/// * `c_timeout` a float, set how long we should try to dial in seconds before timing out
 /// * `rt_ptr` a pointer to a rust runtime previously obtained with init_rust_runtime
 #[no_mangle]
 pub unsafe extern "C" fn dial(
--- a/src/ffi/dial_ffi.rs
+++ b/src/ffi/dial_ffi.rs
@@ -8,6 +8,7 @@ use http::uri::Uri;
 use std::{ptr, time::Duration};
 use tokio::runtime::Runtime;
 use tokio::sync::oneshot;
+use tokio::time::timeout;
 use tracing::Level;
 
 use crate::rpc::dial::{
@@ -103,6 +104,7 @@ fn dial_with_cred(
     let c = if allow_insec { c.allow_downgrade() } else { c };
     Ok(c)
 }
+
 /// Returns a path to a UDS proxy to a robot
 /// # Safety
 ///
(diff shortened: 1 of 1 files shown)


src/ffi/dial_ffi.rs
```diff
@@ -241,7 +252,7 @@ pub unsafe extern "C" fn dial(
     }) {
         Ok(s) => s,
         Err(e) => {
-            log::error!("Error building GRPC proxy reason : {e:?}");
+            log::error!("Error building GRPC proxy reason : {}", e);
```
hexbabe, line 255: More informative error on timeout. Displays "deadline has elapsed" in the error instead of "Elapsed(())". Should be safe as all `Boxed(dyn Error)` have Display implemented.

src/ffi/dial_ffi.rs
```diff
@@ -114,6 +116,7 @@ fn dial_with_cred(
 /// * `c_type` a C-style string representing the type of robot's secret you want to use, set to NULL if you don't need authentication
 /// * `c_payload` a C-style string that is the robot's secret, set to NULL if you don't need authentication
 /// * `c_allow_insecure` a bool, set to true when allowing insecure connection to your robot
+/// * `c_timeout` a float, set how long we should try dial before timing out
```
maximpertsov, line 119: [minor] 

```suggestion
/// * `c_timeout` a float, set how long we should try to dial before timing out
```

maximpertsov (commented):
looks good! one tiny doc comment. also, would it be straightforward to add a unit test for this?

src/ffi/dial_ffi.rs
hexbabe: Would mocking the implementation of `dial_with_cred` to be a time::sleep and making sure it times out a good test?

hexbabe: Wait, or would the test be in Python?

maximpertsov: yeah that totally works, but before that I was wondering how straightforward it would be to call this function in a normal rust unit test context. if we need to build and call in another language then I would say it's not worth it.

maximpertsov (approved):
To be clear, the unit test is optional - it's not super clear to me how simple it is to test functions that are meant to be external: https://github.com/viamrobotics/rust-utils/pull/80#discussion_r1336342174

src/ffi/dial_ffi.rs
hexbabe: So, calling the extern func from within Rust is totally doable. However, in order to call it we have to initialize a new Rust runtime (see `init_rust_runtime` and the `rt_ptr` arg)

Tokio test does not seem to like this

```thread 'test_dial_ffi' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.'```

hexbabe: Maybe this is now at the point where writing this particular test is too big of a hurdle for now ?

maximpertsov: yeah I think so, thank you for trying!

PR #85, opened 2023-12-13
RSDK-6043 Support API key auth in dialdbg

benjirewis:
https://viam.atlassian.net/browse/RSDK-6043

Allow specifying `api-key` auth mechanism and `--entity`. Reports < 1ms average RTT instead of "0ms", which is misleading. Also removes an unnecessary `loop` in webRTC code.

### DIFF
--- a/src/dialdbg/README.md
+++ b/src/dialdbg/README.md
@@ -40,6 +40,11 @@ viam-dialdbg --uri myremote.com --credential mycredential --output ./foo.txt --c
 ```
 Same as above, but only examines WebRTC connection establishment.
 
+``` shell
+viam-dialdbg --uri myremote.com --credential mycredential --output ./foo.txt --credential-type api-key --entity myentity --nogrpc
+```
+Same as above, but uses "api-key" credential type and "myentity" auth entity for "mycredential".
+
 Use `viam-dialdbg --help` for more information.
 
 ## License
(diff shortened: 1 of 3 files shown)


src/dialdbg/README.md
```diff
@@ -40,6 +40,11 @@ viam-dialdbg --uri myremote.com --credential mycredential --output ./foo.txt --c
 ```
 Same as above, but only examines WebRTC connection establishment.
 
+``` shell
+viam-dialdbg --uri myremote.com --credential mycredential --output ./foo.txt --credential-type api-key --entity myentity --nogrpc
```
benjirewis, line 44: I realize this example is a little complex. I'll add an internal wiki page of how to use `viam-dialdbg` and how to interpret its output (I may end up copying some of that page into this `README`, too, for hypothetical external users).

njooma (approved):

cheukt (approved):

stuqdog (approved):

src/rpc/webrtc.rs
```diff
@@ -300,14 +300,12 @@ pub(crate) async fn action_with_timeout<T>(
         let f = f;
     }
 
-    loop {
```
benjirewis, line 303: @stuqdog clippy was erroring because of this `loop`. It looks like this error was only happening with the most recent version of clippy (I wasn't seeing it locally until a ran `rustup update && cargo clippy --all-features`). The linter is complaining because this `loop` is hypothetically unnecessary, as both branches of the `tokio::select!` return a `Result` and end the function, so the loop will never actually iterate.

PR #86, opened 2023-12-21
RSDK-6152 Fix rust-utils release for aarch64

benjirewis:
https://viam.atlassian.net/browse/RSDK-6152

Splits `aarch64` building into its own GH job with a slightly different toolchain setup. Runs the `prepare` job on a buildjet `arm64` machine instead of an `amd64` machine.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,9 +16,9 @@ on:
 jobs:
   prepare:
     if: github.repository_owner == 'viamrobotics'
-    runs-on: [self-hosted, x64]
+    runs-on: [buildjet-8vcpu-ubuntu-2204-arm]
     container:
-      image: ghcr.io/viamrobotics/canon:amd64
+      image: ghcr.io/viamrobotics/canon:arm64
     outputs:
       version: ${{ steps.which_version.outputs.version }}
       sha: ${{ steps.commit.outputs.commit_long_sha }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -16,9 +16,9 @@ on:
 jobs:
   prepare:
     if: github.repository_owner == 'viamrobotics'
-    runs-on: [self-hosted, x64]
+    runs-on: [buildjet-8vcpu-ubuntu-2204-arm]
```
benjirewis, line 19: I stopped short of switching everything over to buildjet, but for some reason the `aarch64` build just didn't work unless _both_ `prepare` and `build_aarch64_linux` ran on an arm64 buildjet machine... I had to switch the container as well to avoid emulation. The other build jobs still run fine it seems.

.github/workflows/release.yml
```diff
@@ -165,8 +163,42 @@ jobs:
           name: builds
           path: builds
 
+  build_aarch64_linux:
```
benjirewis, line 166: This new job is an amalgamation of copy-pasted pieces of `build_linux` and pieces of the micro-RDK's `build-installer-arm64-linux` [job](https://github.com/viamrobotics/micro-rdk/blob/main/.github/workflows/publish.yml#L72-L103). Looks like they ran into a similar issue with `aarch64` and the `ring` crate, and had to push the building out into a separate job; made sense to me to just copy pieces of what they had done. I'll be honest that I don't know _exactly_ what was fixed with these changes, but `aarch64` seems to build fine.

stuqdog (approved):
looks reasonable to me, and it works :)

njooma (approved):

PR #95, opened 2024-02-06
[RSDK-5820] WebRTC connection improvement

npmenard:
This PR make WebRTC connection failure faster when signaling goes wrong instead of waiting for a timeout on is_open. I have also removed AtomicPollableInterrupt and replace it with tokio::sync::* I think these primitive are a bit more flexible and allow for more cooperation between spawned Future. 
Note that it won't attempt to connect again when the connection fails

### DIFF
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -4,6 +4,9 @@ use super::{
     webrtc::{webrtc_action_with_timeout, Options},
 };
 use crate::gen::google;
+use crate::gen::proto::rpc::v1::{
+    auth_service_client::AuthServiceClient, AuthenticateRequest, Credentials,
+};
 use crate::gen::proto::rpc::webrtc::v1::{
     call_response::Stage, call_update_request::Update,
     signaling_service_client::SignalingServiceClient, CallUpdateRequest,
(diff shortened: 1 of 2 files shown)


src/rpc/dial.rs
```diff
@@ -792,16 +790,21 @@ async fn maybe_connect_via_webrtc(
     let sent_done_or_error = Arc::new(AtomicBool::new(false));
     let uuid_lock = Arc::new(RwLock::new("".to_string()));
     let uuid_for_ice_gathering_thread = uuid_lock.clone();
-    let is_open = Arc::new(AtomicBool::new(false));
-    let is_open_read = is_open.clone();
+
+    // Using an mpsc channel to report unrecoverable errors during Signaling, so we
+    // don't have to wait until the timeout expires before giving up on this attempt.
+    // The size of the channel is set to 1 since any error (or success) should terminate the function
+    let (is_open_s, mut is_open_r) = mpsc::channel(1);
```
benjirewis, line 797: If it's an `mpsc` channel with buffer size 1, why not use an [`oneshot` channel](https://docs.rs/tokio/latest/tokio/sync/index.html#oneshot-channel)?

benjirewis (commented):
Generally seems good to me, but @stuqdog will have more context. Just a question.

src/rpc/dial.rs
npmenard: oneshot channels cannot be cloned or reused multiple time (send() takes ownership) We could work around it with Arc and interior mutability but it is a bit less elegant IMO

benjirewis (approved):

stuqdog (approved):
lgtm! thanks for doing this, definitely an improvement

PR #102, opened 2024-03-07
RSDK-4857 musllinux binaries for python sdk

biotinker:


njooma (changes requested):
Looks good. Please rebase from main first

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "viam-rust-utils"
-version = "0.2.1"
+version = "0.2.2"
```
njooma, line 3: Set this back to `0.2.0`

njooma (changes requested):

Cargo.toml
biotinker: Done

njooma (approved):

PR #107, opened 2024-04-02
RSDK-7105 - Fix server streaming

stuqdog:
- Simplifies and corrects the "end of stream" determining logic, removes "eos" arg
- (flyby) parse `next_message_length` as a `usize` initially to reduce cases of `try_into().unwrap()`
- (flyby) remove extraneous `http` import
- (flyby) fix example code credentials template

### DIFF
--- a/src/rpc/client_channel.rs
+++ b/src/rpc/client_channel.rs
@@ -241,6 +241,11 @@ impl WebRTCClientChannel {
                     stream,
                     r#type: Some(Type::Message(RequestMessage {
                         has_message,
+                        // note(ethan): the variable name that used to exist for determining `eos` was
+                        // `it_was_all_a_stream`. Which isn't important at all, but
+                        // `it_was_all_a_stream` is the best variable name I've ever shipped into
+                        // production and it makes me sad to see it go, so I wanted to memorialize
+                        // it somehow!
                         eos,
                         packet_message: Some(PacketMessage {
                             eom: next_message_length == 0 || eos,
--- a/src/rpc/client_channel.rs
+++ b/src/rpc/client_channel.rs
@@ -188,7 +188,6 @@ impl WebRTCClientChannel {
 
     pub(crate) async fn write_message(
         &self,
-        eos: bool,
         stream: Option<Stream>,
         mut data: Vec<u8>,
     ) -> Result<()> {
@@ -222,10 +221,6 @@ impl WebRTCClientChannel {
             to_add_bytes.clone_from_slice(&data[1..5]);
             let mut next_message_length: usize =
                 u32::from_be_bytes(to_add_bytes).try_into().unwrap();
-            // if this is all streaming calls we need to tell the server when we're done with
-            // the stream, otherwise neither side will know we're done, trailers will never be
-            // sent/processed, and we'll hang on the stream.
-            let it_was_all_a_stream = next_message_length + 5 < data.len();
 
             data = data.split_off(5);
             // we need an internal loop because a single message may be longer than the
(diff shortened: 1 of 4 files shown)


src/rpc/client_channel.rs
```diff
@@ -205,13 +204,6 @@ impl WebRTCClientChannel {
         // header math in the process.
 
         let mut to_add_bytes = [0u8; 4];
-        // 1-5 because those are the length header bytes for gRPC
-        to_add_bytes.clone_from_slice(&data[1..5]);
-        let mut next_message_length = u32::from_be_bytes(to_add_bytes);
-        // if this is all streaming calls we need to tell the server when we're done with
-        // the stream, otherwise neither side will know we're done, trailers will never be
-        // sent/processed, and we'll hang on the strream.
-        let it_was_all_a_stream = usize::try_from(next_message_length).unwrap() + 5 < data.len();
```
stuqdog, line 214: All of this logic to determine whether or not we were in a stream was only done once, at the top level loop. This made sense insofar as it would accurately tell us if we were in a streaming scenario (a fact that wouldn't change), but it wasn't ideal for determining when we were at end of stream (which we were using it for).

src/rpc/client_channel.rs
```diff
@@ -236,24 +232,18 @@ impl WebRTCClientChannel {
...
                     r#type: Some(Type::Message(RequestMessage {
                         has_message,
-                        eos: if !remaining.is_empty() {
-                            // stream definitely isn't done if there's more to send
-                            false
-                        } else {
-                            // if we intentionally sent an eos or the http request was inferrably
-                            // a stream
-                            eos || it_was_all_a_stream
-                        },
```
stuqdog, line 254: Funny enough, we had a bit of a "two wrongs _do_ make a right" scenario here. We shouldn't be saying "stream done" every time we're in a stream (which we were doing before), but this actually led to client streaming working correctly instead because we were just defaulting to saying it was _always_ eos, which seemed to be fine.

The issue seems to be that in unary calls, `eos` (as far as I can tell) is meaningless so the fact that we were always saying "false" for eos in those cases (because `eos` was being passed as false, and because it definitely wasn't a stream so `it_was_all_a_stream` evaluated to false) didn't really matter. In client streaming cases, because we weren't ever re-evaluating the `it_was_all_a_stream` value, we were actually passing "true" here at all times! Thus, everything worked out okay.

 But in server streaming cases, `it_was_all_a_stream` was evaluating to false and so we were saying `eos: false`. This is exactly the same as in the unary case, which worked perfectly normally. My theory is that on the server side, `eos` is evaluated/matters in all streaming call cases (thus causing issues with our server streaming request despite the fact that the request is unary), but never in unary call cases (thus allowing unary requests to succeed despite the fact that `eos` was always false).

src/rpc/client_channel.rs
```diff
@@ -236,24 +232,18 @@ impl WebRTCClientChannel {
             loop {
                 let split_at = MAX_REQUEST_MESSAGE_PACKET_DATA_SIZE
                     .min(data.len())
-                    .min(usize::try_from(next_message_length).unwrap());
+                    .min(next_message_length);
                 let (to_send, remaining) = data.split_at(split_at);
-                next_message_length -= u32::try_from(split_at).unwrap();
+                next_message_length -= split_at;
                 let stream = stream.clone();
+                let eos = remaining.is_empty();
```
stuqdog, line 239: We got bit here a bit by trying to be too close to `Go`. The Go logic doesn't seem to have a way of reliably saying that the stream is done, hence taking `eos` as an argument. We initially copied that in Rust, but it turns out we _do_ have a good way to determine if the stream is done or not, since we receive a complete vec of all the calls. Thus we can just check for that `vec` to be empty, removing the need for an `eos` value to ever be passed and reducing the risk of it being passed incorrectly.

Worth noting, the case where there is meaningful latency delay between messages being sent on the client side in a bidi streaming context remains a little uncertain. We seem to consistently receive all messages in a single vec, and thus can reliably infer when `eos` occurs. I'm not certain what will happen in cases where there's a heavy delay (is there blocking somewhere in rust to collect all the stream messages? will we end up making multiple `call` calls under the hood?), in part because (to my knowledge) we don't have any APIs that work that way yet. But if and when we do, some revisiting may be necessary. There's a comment to this effect in the file already.

src/rpc/client_channel.rs
```diff
@@ -236,24 +232,23 @@ impl WebRTCClientChannel {
...
-                        } else {
-                            // if we intentionally sent an eos or the http request was inferrably
-                            // a stream
-                            eos || it_was_all_a_stream
-                        },
+                        // note(ethan): the variable name that used to exist for determining `eos` was
+                        // `it_was_all_a_stream`. Which isn't important at all, but
+                        // `it_was_all_a_stream` is the best variable name I've ever shipped into
+                        // production and it makes me sad to see it go, so I wanted to memorialize
+                        // it somehow!
```
stuqdog, line 248: 😭

stuqdog:
(note to reviewers): I'm still doing a liiittle bit more testing as I consider more cases so it's possible there will be future changes based on that. But, I have decently high confidence that this is correct and didn't want to prevent review from happening. If anything meaningful _does_ change, I will be sure to ping reviewers and flag it.

njooma (approved):

PR #109, opened 2024-04-11
RSDK-5993 - fallback to non-mdns uri when auth fails

stuqdog:
#### Major changes
When attempting to connect with an mDNS uri fails, we now fallthrough and attempt a standard connection.

Attempts to auth to a robot with an org level API key (or presumably any level other than the robot itself) fail when connecting with mDNS. Previously we would just fail; we've now added explicit fallback logic.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -3775,7 +3775,7 @@ dependencies = [
 
 [[package]]
 name = "viam-rust-utils"
-version = "0.2.0"
+version = "0.2.4"
 dependencies = [
  "anyhow",
  "base64 0.13.1",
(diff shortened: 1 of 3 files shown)


njooma (approved):

purplenicole730 (commented):
I admittedly have no idea what I read (multiple times) 😂 
While I don't really understand the language or the background, it does look like it does what your description set out to do. Thought to comment even if I can't really "approve".

stuqdog:
> I admittedly have no idea what I read (multiple times) 😂 While I don't really understand the language or the background, it does look like it does what your description set out to do. Thought to comment even if I can't really "approve".

@purplenicole730 I'd be happy to grab a side room and chat offline for a couple minutes if you'd like me to go over/explain anything!

PR #115, opened 2024-05-24
RSDK-6283 - parallel dial attempts

stuqdog:


src/rpc/dial.rs
```diff
@@ -459,7 +453,7 @@ impl<T: AuthMethod> DialBuilder<T> {
                     let mut uri_parts = uri.clone().into_parts();
                     uri_parts.scheme = Some(Scheme::HTTP);
                     let uri = Uri::from_parts(uri_parts)?;
-                    Channel::builder(uri.clone()).connect().await?
+                    Channel::builder(uri).connect().await?
```
stuqdog, line 456: (flyby) this was an unnecessary clone

src/rpc/dial.rs
```diff
@@ -470,6 +464,20 @@ impl<T: AuthMethod> DialBuilder<T> {
...
+            config: DialOptions {
+                credentials: None,
+                webrtc_options: self.config.webrtc_options.clone(),
+                uri: self.duplicate_uri(),
+                disable_mdns: self.config.disable_mdns,
+                allow_downgrade: self.config.allow_downgrade,
+                insecure: self.config.insecure,
+            },
+        }
+    }
```
stuqdog, line 479: the `connect()` method consumes a `DialBuilder`, so we need to be able to clone them in order to allow parallel attempts. Incidentally, this allows us to clean up a bit some of the dial logic below.

src/rpc/dial.rs
```diff
@@ -550,21 +556,65 @@ impl DialBuilder<WithoutCredentials> {
...
-            None => {
-                return Err(anyhow::anyhow!(
-                    "Attempting to connect but there was no uri"
-                ))
+        let original_uri = self.duplicate_uri().ok_or(anyhow::anyhow!(
+            "Attempting to connect but there was no uri"
+        ))?;
+        let original_uri2 = duplicate_uri(&original_uri).ok_or(anyhow::anyhow!(
+            "Attempting to connect but there was no uri"
+        ))?;
```
stuqdog, line 579: Slightly annoying that this is duplicative. `original_uri2` will never fail (if it were to do so, `original_uri` would have failed before so we _could_ `unwrap()` rather than the `ok_or` block, but I think adding `unwrap`s is a bit uglier/less idiomatic than the minorly duplicative error handling here.

purplenicole730 (commented):
I read through it, but I would be lying if I said I understood most of it. Some nits, but I'll leave actual approval to those who understand what's going on 😅

njooma (approved):

PR #119, opened 2024-06-20
RSDK-7282 Log ICE event timings/stats

benjirewis:
[RSDK-7282](https://viam.atlassian.net/browse/RSDK-7282)

Adds info-level logs for:
- Gathering of local ICE candidates
- Reception of remote ICE candidates
- Statistics of caller updates for this connection establishment
     - Number of updates
     - Average duration of update
     - Max duration of update

[RSDK-7282]: https://viam.atlassian.net/browse/RSDK-7282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -959,7 +959,7 @@ async fn maybe_connect_via_webrtc(
                 Box::pin(async move {
                     if state == RTCIceConnectionState::Completed {
                         let caller_update_stats_inner = caller_update_stats.lock().unwrap();
-                        log::info!("{}", caller_update_stats_inner);
+                        log::debug!("{}", caller_update_stats_inner);
                     }
                 })
             },
@@ -1005,7 +1005,7 @@ async fn maybe_connect_via_webrtc(
                     let mut signaling_client = SignalingServiceClient::new(channel.clone());
                     match ice_candidate {
                         Some(ice_candidate) => {
-                            log::info!("Gathered local candidate of {}", ice_candidate);
+                            log::debug!("Gathered local candidate of {}", ice_candidate);
                             if sent_done_or_error.load(Ordering::Acquire) {
                                 return;
                             }
@@ -1197,7 +1197,7 @@ async fn maybe_connect_via_webrtc(
                                     break;
                                 }
                             };
-                            log::info!("Received remote ICE candidate of {:#?}", candidate);
+                            log::debug!("Received remote ICE candidate of {:#?}", candidate);
                             if let Err(e) = client_channel
                                 .base_channel
                                 .peer_connection
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -884,7 +884,7 @@ impl fmt::Display for CallerUpdateStats {
         let average_duration = &self.total_duration.as_millis() / &self.count;
         writeln!(
             f,
-            "Caller update statistics: num_updates {}, average_duration: {}ms, max_duration: {}ms",
+            "Caller update statistics: num_updates: {}, average_duration: {}ms, max_duration: {}ms",
             &self.count,
             average_duration,
             &self.max_duration.as_millis()
@@ -1005,7 +1005,7 @@ async fn maybe_connect_via_webrtc(
                     let mut signaling_client = SignalingServiceClient::new(channel.clone());
                     match ice_candidate {
                         Some(ice_candidate) => {
-                            log::debug!("Gathered local candidate of {}", ice_candidate);
+                            log::debug!("Gathered local candidate of {ice_candidate}");
                             if sent_done_or_error.load(Ordering::Acquire) {
                                 return;
                             }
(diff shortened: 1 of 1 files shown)


src/rpc/dial.rs
```diff
@@ -925,6 +950,20 @@ async fn maybe_connect_via_webrtc(
         let exchange_done = exchange_done.clone();
 
         let on_local_ice_candidate_failure = is_open_s.clone();
+
+        let caller_update_stats = caller_update_stats.clone();
+        let caller_update_stats2 = caller_update_stats.clone();
```
stuqdog, line 955: I believe `caller_update_stats2` is not necessary. You should be able to clone `caller_update_stats` again because ownership of `caller_update_stats` isn't moving.

src/rpc/dial.rs
```diff
@@ -988,6 +1030,16 @@ async fn maybe_connect_via_webrtc(
...
+                                    let mut caller_update_stats_inner =
+                                        caller_update_stats.lock().unwrap();
+                                    caller_update_stats_inner.count += 1;
+                                    let call_update_duration = call_update_start.elapsed();
+                                    if call_update_duration > caller_update_stats_inner.max_duration
+                                    {
+                                        caller_update_stats_inner.max_duration =
+                                            call_update_duration;
+                                    }
+                                    caller_update_stats_inner.total_duration = call_update_duration;
```
stuqdog, line 1042: maybe I'm not following correctly, but it looks to me that `total_duration` is updated with each ice candidate, and is updated to be the duration involved for that specific candidate. If that's correct, maybe we should change this to `current_duration`? `max_duration` to me implies the duration spent processing _all_ ice candidates.

stuqdog: Ahh based on the reference to average duration in the `Display` implementation it looks like you are looking for the total duration to process all candidates. I _think_ that's not what we're getting here, since `total_duration` is updated based on `call_update_start` but `call_update_start` is redefined with each candidate. Should `call_update_start` be defined outside of `on_ice_candidate`? I'm thinking define it outside as a mutable `Option` type and then update it in `on_ice_candidate` only if it's a `None`, to make sure it stays stable within candidates while still accurately reflecting when you first started processing ice candidates.

src/rpc/dial.rs
```diff
@@ -965,6 +1005,7 @@ async fn maybe_connect_via_webrtc(
                     let mut signaling_client = SignalingServiceClient::new(channel.clone());
                     match ice_candidate {
                         Some(ice_candidate) => {
+                            log::debug!("Gathered local candidate of {}", ice_candidate);
```
stuqdog, line 1008: (minor, opt) we aren't super consistent about it, but in general I think `"Gathered local candidate of {ice_candidate}"` is cleaner than `"Gathered local candidate of {}", ice_candidate`. Similar comment with `log` statement below. No need to change if you disagree!

stuqdog (commented):
Nice, looks good. Just a couple minor things and some uncertainty around the `total_duration` logic.

src/rpc/dial.rs
```diff
@@ -869,6 +872,27 @@ async fn send_done_once(
...
+    total_duration: Duration,
+    max_duration: Duration,
+}
+
+impl fmt::Display for CallerUpdateStats {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let average_duration = &self.total_duration.as_millis() / &self.count;
+        writeln!(
+            f,
+            "Caller update statistics: num_updates {}, average_duration: {}ms, max_duration: {}ms",
```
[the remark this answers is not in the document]
benjirewis: Ah yep good catch!

src/rpc/dial.rs
benjirewis: Hmm I do think `caller_update_stats` gets moved in the `on_ice_connection_state_change` call below. If I try to reuse the value (even to clone it) in `on_ice_candidate` below _that_, the compiler complains that the value has already been moved above.

src/rpc/dial.rs
benjirewis: 😮‍💨 entirely right; this should be `caller_update_stats_inner.total_duration += call_update_duration`. I _think_ that should fix things here, but let me know if you disagree.

src/rpc/dial.rs
benjirewis: Yep can do! I always forget you can put the variable directly in the braces.

src/rpc/dial.rs
stuqdog: Ah yep, I think that would work! Didn't realize you could add durations, that's neat :)

src/rpc/dial.rs
stuqdog: oops! borrow semantics are confusing, disregard my above comment then :)

stuqdog (approved):

cheukt (approved):

PR #121, opened 2024-06-24
RSDK-7986 - stop logging noisy errors

stuqdog:
We were generating noisy errors in `rust-utils` occasionally when trying to connect. This seemed to be the result of parallel dialing, but it happened non-deterministically. While the errors were basically harmless (one branch always succeeds and so a connection is always still formed), they were noisy and scary-looking (it could be easy to miss that a connection was in fact formed successfully and that any calls were made successfully).

Some digging revealed that the cause of this problem was a failure to receive updates from the signaling service, meaning we never set a `UUID` and so sent an empty string as our UUID. Thus, we can avoid the noisy errors by checking if the UUID is empty before sending an update request and just not sending if so. The exact cause of the problem is unknown to me at this time; I'd love to dig into it more but given that this issue doesn't seem to be actually affecting performance or functionality in any way, it's probably not worth the effort at this time. Very happy to revisit when that changes (or if anyone disagrees)!

cc @benjirewis, who helpfully noticed and flagged this issue to begin with.

### DIFF
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -962,6 +962,25 @@ async fn maybe_connect_via_webrtc(
                     }
 
                     let uuid = uuid_lock.read().unwrap().to_string();
+                    // Note(ethan): for reasons that aren't entirely clear to me, parallel dialing
+                    // occasionally causes us to not receive a signaling client response when
+                    // trying to establish a connection. This results in noisy error messages that
+                    // fortunately are harmless (this problem seems to only ever affect one branch
+                    // of the parallel dial, so we still end up with a successful connection).
+                    // By checking if the `uuid` is empty, we can tell if we're in such a case and
+                    // exit out before it results in logging noisy error messages.
+                    //
+                    // It would be lovely to understand this problem better, but given that it's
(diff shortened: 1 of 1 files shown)


benjirewis:
> one branch always succeeds and so a connection is always still formed

Is it possible for both connections to succeed and us to "leak" one of them? We had that issue in go.

stuqdog:
> > one branch always succeeds and so a connection is always still formed
> 
> Is it possible for both connections to succeed and us to "leak" one of them? We had that issue in go.

It shouldn't be, at least as I understand it. Rust drops the unused connection, which should be closing the data channel. Some rudimentary testing seems to support this, but I confess I'm not 100% confident here because I'm not sure what exactly I should be looking for as signs of a leak. How did y'all identify the leak?

benjirewis:
> Rust drops the unused connection, which should be closing the data channel.

Cool; I think that should be fine. IIRC the mechanics of goroutines were kind of limiting us and we had a difficult time ensuring the second goroutine to establish a connection both halted and closed its connection.

njooma (approved):

PR #122, opened 2024-07-01
chore(release): add ARM musl build targets

HipsterBrown:
Second attempt to add `musl` support for aarch64/arm build targets to support in-progress Home Assistant core integration, which requires the Python SDK to run on more platforms. See #112 for more information.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -140,9 +140,21 @@ jobs:
             platform: musllinux_x86_64
             image: alpine
             runs-on: buildjet-4vcpu-ubuntu-2204
+          - target: aarch64-unknown-linux-musl
+            platform: musllinux_aarch64
+            image: ghcr.io/cross-rs/aarch64-unknown-linux-musl:main
+            runs-on: buildjet-4vcpu-ubuntu-2204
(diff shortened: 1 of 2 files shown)


purplenicole730 (approved):

.github/workflows/release.yml
```diff
@@ -140,10 +140,22 @@ jobs:
...
+            image: ghcr.io/cross-rs/aarch64-unknown-linux-musl:main
+            runs-on: buildjet-4vcpu-ubuntu-2204
           - target: arm-unknown-linux-gnueabihf
             platform: linux_armv6l
             image: ghcr.io/cross-rs/arm-unknown-linux-gnueabihf:main
             runs-on: buildjet-4vcpu-ubuntu-2204
+          - target: arm-unknown-linux-musleabihf
+            platform: linux_armv6l
+            image: ghcr.io/cross-rs/arm-unknown-linux-musleabihf:main
+            runs-on: buildjet-4vcpu-ubuntu-2204
```
njooma, line 154: @HipsterBrown This will actually cause a collision and only a single `linux_armv6l` build will be produced. The platform name must be different, and should probably be `musllinux_armv6l`

PR #125, opened 2024-07-10
chore(release): update musl armv6l, armv7l platform name to musllinux

HipsterBrown:
Fixes https://github.com/viamrobotics/rust-utils/pull/122#discussion_r1672629323

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -138,11 +138,11 @@ jobs:
             image: ghcr.io/cross-rs/arm-unknown-linux-gnueabihf:main
             runs-on: buildjet-4vcpu-ubuntu-2204
           - target: arm-unknown-linux-musleabihf
-            platform: linux_armv6l
+            platform: musllinux_armv6l
             image: ghcr.io/cross-rs/arm-unknown-linux-musleabihf:main
             runs-on: buildjet-4vcpu-ubuntu-2204
           - target: armv7-unknown-linux-musleabihf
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -138,11 +138,11 @@ jobs:
             image: ghcr.io/cross-rs/arm-unknown-linux-gnueabihf:main
             runs-on: buildjet-4vcpu-ubuntu-2204
           - target: arm-unknown-linux-musleabihf
-            platform: linux_armv6l
+            platform: muslinux_armv6l
```
njooma, line 141: Can we add the second `l` here? `musllinux` pls

.github/workflows/release.yml
```diff
@@ -138,11 +138,11 @@ jobs:
...
             image: ghcr.io/cross-rs/arm-unknown-linux-musleabihf:main
             runs-on: buildjet-4vcpu-ubuntu-2204
           - target: armv7-unknown-linux-musleabihf
-            platform: linux_armv7l
+            platform: muslinux_armv7l
```
njooma, line 145: And here

njooma (changes requested):

.github/workflows/release.yml
njooma: (Because python likes the double `l`: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#musllinux)

PR #128, opened 2024-08-14
RSDK-8469 - Fix offline dial

stuqdog:
Adds support for loopback candidates in webRTC, allowing us to establish webRTC connections via mDNS in offline scenarios. Previously, offline connectivity only worked via gRPC in rust-utils.

Note that this relies on fixes to the upstream `webrtc-rs` crate which have not yet been merged. As such, we have to point to my branch of the `webrtc` crate in `Cargo.toml`. It might be good to consider whether we merge this to a separate branch than `main`, or hold off on merging until upstream fixes are merged. Either way, it seemed good to have this up at least so people can look and use as necessary.

The upstream fix, if anyone cares to see, is located at https://github.com/webrtc-rs/webrtc/pull/603

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,45 +151,18 @@ checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
 
 [[package]]
 name = "asn1-rs"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


purplenicole730 (approved):
Looks good from what I could tell, I have to admit there's a knowledge gap

stuqdog:
> LGTM from me too, maybe worth waiting for it to land in the mainline release? I'm seeing that your PR was merged yesterday

Yeah I'm gonna update to point to that commit in their github repo, but probably won't wait until they do an actual release (those tend to be pretty infrequent and I'd like to get this out sooner rather than later). 

We can always update to point to the official release once it goes out.

PR #140, opened 2025-01-24
RSDK-9300 - make mdns query more generic over local-ness of candidate

stuqdog:
Fixes issue where `rust-utils` was unable to find an mDNS candidate for micro-rdk robots.

Tested locally: before change, was unable to connect to a native micro-rdk process with signaling disabled. After change, I was able to connect to the same micro-rdk process whether signaling was enabled or not. Also tested to ensure no regression in connectivity to a locally running `viam-server` process, both with wifi on and with wifi off.

cc @acmorrow

### DIFF
--- a/src/rpc/dial.rs
+++ b/src/rpc/dial.rs
@@ -329,13 +329,8 @@ impl<T: AuthMethod> DialBuilder<T> {
         let mut resp: Option<Response> = None;
         for ipv4 in addresses {
             for candidate in candidates {
-                let mut addr_to_send = "".to_string();
-                addr_to_send.push_str(candidate.as_str());
-                addr_to_send.push('.');
-                addr_to_send.push_str(VIAM_MDNS_SERVICE_NAME);
-
                 let discovery = discover::interface_with_loopback(
-                    addr_to_send,
+                    VIAM_MDNS_SERVICE_NAME,
                     Duration::from_millis(250),
                     ipv4,
                 )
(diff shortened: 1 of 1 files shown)


src/rpc/dial.rs
```diff
@@ -400,7 +403,7 @@ impl<T: AuthMethod> DialBuilder<T> {
 
         let ifaces: HashMap<&str, Vec<&IpAddr>> =
             ifaces.iter().fold(HashMap::new(), |mut map, (k, v)| {
-                map.entry(k).or_insert(vec![]).push(v);
+                map.entry(k).or_default().push(v);
```
stuqdog, line 406: flyby

src/rpc/dial.rs
```diff
@@ -329,13 +329,8 @@ impl<T: AuthMethod> DialBuilder<T> {
...
         for ipv4 in addresses {
             for candidate in candidates {
-                let mut addr_to_send = "".to_string();
-                addr_to_send.push_str(candidate.as_str());
-                addr_to_send.push('.');
-                addr_to_send.push_str(VIAM_MDNS_SERVICE_NAME);
-
                 let discovery = discover::interface_with_loopback(
-                    addr_to_send,
+                    VIAM_MDNS_SERVICE_NAME,
```
stuqdog, line 333: See added comment re: the localness of machine URIs to understand why this is important. Basically, we were trying to discover candidates with a fully qualified name. but, the fully qualified name we were passing was non-local and, for reasons not entirely clear to me, the only candidates we could find from a micro-rdk process were the ones with a local name. By making our search more generic we can find everything that looks Viam-y, and then narrow it down based on the actual name of the machine.

PR #145, opened 2025-04-14
RSDK-10366 - rust utils for windows

stuqdog:
Support rust-utils on windows through use of a TCP proxy.

Tested on a windows machine, was able to connect through `rust-utils` via webRTC successfully.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,6 @@
 name: Build + draft release
 
 on:
-  pull_request:
   workflow_dispatch:
     inputs:
       version:
@@ -16,7 +15,7 @@ on:
         - nobump
 jobs:
   prepare:
-    #if: github.repository_owner == 'viamrobotics'
+    if: github.repository_owner == 'viamrobotics'
     runs-on: buildjet-8vcpu-ubuntu-2204-arm
     outputs:
       version: ${{ steps.which_version.outputs.version }}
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -201,8 +201,45 @@ jobs:
...
+          cargo build --release --target=${{ matrix.target }}
+      - name: Copy
+        run: |
+          cp target/${{ matrix.target }}/release/viam_rust_utils.dll builds/libviam_rust_utils-${{ matrix.platform }}.dll
+          cp target/${{ matrix.target }}/release/viam_rust_utils.lib builds/libviam_rust_utils-${{ matrix.platform }}.lib
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: builds-${{ matrix.platform }}
+          path: builds/*
```
stuqdog, line 239: Made reasonable modifications based on research/testing locally, but this is all untested as a CI workflow; didn't want to mess around with trying to make actual builds when running from a PR on my fork, and I figure it makes some sense to push the release stuff specifics to a separate PR. Happy to discuss if there's any disagreement or concern.

njooma: Curious to see if the cross-compile works between aarch64 and x86 on windows

njooma (approved):

.github/workflows/release.yml
stuqdog: We'll find out but I suspect it will, given that we're setting up the rust toolchain for the specific architecture and `cargo` supports cross-compilation out of the box.

PR #146, opened 2025-04-14
Update license_finder.yml

stuqdog:
This job has been failing because we don't have `Unicode-3.0` on the list of permitted licenses. So, let's add it.


stuqdog:
probably better to update `doc/dependency_decisions.yml` instead.

PR #148, opened 2025-04-17
Update release.yml

stuqdog:
Fixes windows bulding in CI. tested, works successfully (I created a new draft release and promptly deleted).

Note that I removed the arm build; it was failing due to an issue with the ring crate (see https://github.com/briansmith/ring/issues/2135). this is solvable but will take some time; ignoring for now since arm windows isn't really a thing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "viam-rust-utils"
-version = "0.3.0"
+version = "0.4.0"
 edition = "2021"
 license = "Apache-2.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "viam-rust-utils"
-version = "0.3.0"
+version = "0.2.17"
```
stuqdog, line 3: this got accidentally bumped to `0.3.0` in testing. I'm just setting it back to where it's supposed to be.

PR #170, opened 2026-03-20
[RSDK-13658] Add force relay and force p2p dial flags

danielbotros:
## Description

[RSDK-13658](https://viam.atlassian.net/browse/RSDK-13658) figure out why metrics change in coturn caused rustutils failure.

Part of changes investigating why `rust-utils` panics when dialing against Coturn `latest`. Adds diagnostic tooling to reproduce and test signaling/ICE scenarios. (The underlying stun panic is not fixed in this PR). These changes can be used for future testing of `rust-utils` dial code against all signaling paths / coturn changes.

* Add a suite of `viam-dialdbg` flags / WebRTC `DialOptions` for targeted WebRTC/TURN testing:
    * `--force-relay` sets ICE policy to only use relay candidates
    * `--force-p2p` strips TURN servers from the ICE config so only host/srflx candidates are used
    * `--signaling-server` overrides the app/signaling address returned by `infer_remote_uri_from_authority`
    * `--disable-mdns` causes mDNS paths to return early
    * `--turn-uri <uri>` filters the signaling server's TURN list to only the server matching the given URI (compared by scheme, host, port, and transport — defaulting transport to UDP if omitted). Example: `turn:turn.viam.com:443`
* Update `stats.rs` report to print the nominated ICE candidate pair (looked up via `CandidatePair.nominated`) and include `candidate_type` in the output, followed by all gathered candidates.
* Add a test script (`src/dialdbg/test-relay-and-p2p-dial-opts.sh`) that runs six end-to-end scenarios covering baseline, `--force-relay`, `--turn-uri`, `--force-p2p`, and an expected-failure case.
* Fix panic in `stats.rs`: use `cand.timestamp.elapsed()` instead of `now.duration_since(cand.timestamp)`, which panics when the candidate timestamp is newer than `now`

## Testing

- [x] `cargo build --features dialdbg` compiles cleanly
- [x] `viam-dialdbg --force-relay` establishes a relay-only connection
- [x] `viam-dialdbg --force-p2p` connects without any TURN relay
- [x] `viam-dialdbg --turn-uri <uri>` with `--force-relay` routes through the matched TURN server
- [x] `viam-dialdbg --turn-uri turn:notexist.example.com:3478` with `--force-relay` fails (no relay candidates)
- [x] `viam-dialdbg --signaling-server <addr>` dials against a specific app instance
- [x] `--force-relay` and `--force-p2p` are mutually exclusive 
### Validated Test Suite

[...]

PR #176, opened 2026-04-28
[RSDK-13658] Plumb force_relay/force_p2p/turn_uri through C FFI

danielbotros:
## Summary
- Plumbs the `DialBuilder` options landed in #170 (`force_relay`, `force_p2p`, `turn_uri`) through the C FFI as an opaque, ABI-stable options handle; new fields in the future require only a new setter, never a header change.
- `viam_dial` (7-arg) is left untouched. Non-Rust consumers (C++ SDK) see no breaking change. It is now `#[deprecated]` in favor of `viam_dial_with_opts`.
- Consolidates `viam_dial`, the deprecated `dial`, and the new `viam_dial_with_opts` onto a shared private `dial_impl` so they can't drift.
- Consumed by viamrobotics/viam-python-sdk#1208.
- New FFI surface:
  ```c
  void *viam_dial_opts_new(void);
  void  viam_dial_opts_free(void *opts);
  void  viam_dial_opts_set_force_relay(void *opts, bool value);
  void  viam_dial_opts_set_force_p2p(void *opts, bool value);
  void  viam_dial_opts_set_turn_uri(void *opts, const char *value);
  char *viam_dial_with_opts(..., const void *opts);  // NULL opts = defaults
  ```
- The internal `DialOpts` struct never appears in the generated header; cbindgen sees only `void *`. Future field additions don't break the C ABI.

## Test plan
- [x] `cargo build --release` + `cargo clippy --all-features` clean
- [x] `cargo test --release --lib` passes (4/4)
- [x] `nm` confirms exports: `_viam_dial_opts_new/_free/_set_*`, `_viam_dial_with_opts`, `_viam_dial` (deprecated), `_dial` (deprecated)
- [x] `cbindgen` regenerates a clean header. Every opts param is `void *`, the `DialOpts` Rust struct never appears
- [x] C test harness compiles+links: exercises 7-arg `viam_dial`, opts lifecycle, all setters (NULL/empty/value paths), and `viam_dial_with_opts`. `-Wall -Wextra` clean, links against the dylib
- [x] (manual) coturn dial verifying `force_relay` pins to relay candidates and `force_p2p` strips them — pending with the paired python-sdk PR


stuqdog (changes requested):
The logic changes all look reasonable but I'm pretty sure this will break existing C++ SDK instances. When C++ builds it looks for an existing rust-utils instance and if it doesn't find one it just downloads the latest release. 

One possible fix would be to update the cbindgen process to generate a function overload signature for `viam_dial` that uses the old signature and puts default values in. Happy to chat more in person if you have questions or thoughts!

danielbotros:
> The logic changes all look reasonable but I'm pretty sure this will break existing C++ SDK instances. When C++ builds it looks for an existing rust-utils instance and if it doesn't find one it just downloads the latest release.
> 
> One possible fix would be to update the cbindgen process to generate a function overload signature for `viam_dial` that uses the old signature and puts default values in. Happy to chat more in person if you have questions or thoughts!

I went ahead and changed the cbindgen config to emit a C++ only header that fills in default values for the last 3 new dial options args so the C++ SDK can compile against the 10-arg header. 

This requires a small follow up PR to the C++ SDK to update the `viam_dial` stub to 10-args for machines on a non-x86_64 Windows platform to compile against since they don't pull a rust-utils binary and can't dial but can still interface modules. Pretty sure this a breaking change for them, if anyone is even running that kind of setup...

stuqdog (approved):
nice, this looks good to me!

PR #177, opened 2026-05-04
RSDK-13852 - fix mdns connection failures

stuqdog:
Co-written with claude code. 

Note that this is blocked on https://github.com/stuqdog/mdns/pull/2 being released, test failures until that time are expected.

Changes: 
- when creating a channel over loopback, don't try TLS since viam-server advertises on localhost without TLS.
- (Temporarily) stop trying to verify TLS certificates, because there are existing certs that are only provided with the leaf cert, which causes authentication to fail. On May 22nd this should be resolved, at which point we will revert.

Testing:
- Confirmed that online connection via both mDNS and regular webRTC worked, and that offline connection via loopback mDNS was successful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ futures-util = "0.3"
 http = "0.2.7"
 http-body = {version = "0.4.4"}
 hyper = { version = "0.14.20", features = ["full"]  }
+hyper-rustls = { version = "0.24", features = ["http2"] }
+rustls = { version = "0.21", features = ["dangerous_configuration"] }
 interceptor = "0.12.0"
 libc = {version = "0.2"}
 local-ip-address = "0.5.5"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -61,6 +63,11 @@ viam-mdns = "3.0.1"
 webpki-roots = "0.21.1"
 webrtc = "0.12.0"
 
+# Override viam-mdns with the local development version while the macOS fix is
+# not yet published.  Remove this section before releasing a new crate version.
+[patch.crates-io]
+viam-mdns = { path = "../mdns" }
```
stuqdog: This will be removed before merging.

stuqdog:
@viamrobotics/netcode adding you all here because this work is veering into netcode-y territory. Happy to chat with y'all about it! Also note that https://github.com/stuqdog/mdns/pull/2 is a necessary precursor here.

cheukt:
seems fine to me, @danielbotros will probably have the most context

src/rpc/dial.rs
```diff
@@ -443,6 +532,18 @@ impl<T: AuthMethod> DialBuilder<T> {
 
         let ifaces = list_afinet_netifas().ok()?;
 
+        // Collect all local IPv4 addresses so we can validate mDNS response IPs.
+        // viam-server may advertise a stale IP (e.g. a WiFi address from when it started,
+        // now unreachable because WiFi is off). By only accepting non-loopback IPs that are
+        // currently present on a local interface, we avoid picking an unreachable address.
```
danielbotros: nit: comment is a little confusing since we don't filter out loopback IPs here, only at the use-site

danielbotros (approved):
LGTM, thanks!

PR #184, opened 2026-07-01
RSDK-654 - refactor message handling for streaming

stuqdog:
Refactors message handling to send messages asynchronously as soon as they come in, defer sending EOS until client messages are demonstrably completed, and only as its own frame.

### DIFF
--- a/src/rpc/client_channel.rs
+++ b/src/rpc/client_channel.rs
@@ -186,90 +186,65 @@ impl WebRTCClientChannel {
         self.send(&header_vec).await
     }
 
-    pub(crate) async fn write_message(
-        &self,
-        stream: Option<Stream>,
-        mut data: Vec<u8>,
-    ) -> Result<()> {
-        // even if no meaningful data, any actual message will include at least frame header bytes
-        let has_message = !data.is_empty();
+    /// Sends a single complete gRPC-framed message (1-byte compression flag + 4-byte
+    /// big-endian length + payload), packetizing the payload across multiple Requests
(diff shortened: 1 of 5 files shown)


acmorrow:
@stuqdog - I haven't dug into the code, but I have pulled this PR locally and confirmed that in my repro environment, I no longer see the arm wait until all the points have been sent on the stream. So this looks promising!

src/rpc/client_channel.rs
```diff
@@ -186,92 +186,67 @@ impl WebRTCClientChannel {
...
+            if eom {
                 break;
             }
         }
         Ok(())
     }
 
+    /// Signals the end of the client's send stream (the gRPC CloseSend equivalent).
+    // note(ethan): the variable that used to carry this signal was named
+    // `it_was_all_a_stream` — the best variable name I've ever shipped into production.
```
acmorrow, line 236: I love that this is a memorial to a memorial.

acmorrow (approved):
I took a pass through this, but it has been over a year since I wrote any production rust. You might want @npmenard to also take a look, especially since he did the BiDi work in the MicroRDK and knows the async rust gRPC interaction quite well.

npmenard (approved):
LGTM

src/rpc/client_channel.rs
stuqdog: I just couldn't let it go haha

acmorrow: I understand completely! Long ago, in a closed source system, I managed to slip in the following haiku, which I now have an opportunity to bring back to life:
```
Exception rising
On the signal handler stack
Portending sorrow
```

PR #186, opened 2026-07-10
align rotation-matrix FFI to row-major convention

JosephBorodach:
Applying the analogous fix to what shipped in rdk#6193 (https://github.com/viamrobotics/rdk/pull/6193).

Two motivations: 
- The Python SDK advertises itself as using row-major, but calls into rust-utils which uses column-major. The two disagree. Example comment from python sdk of incorrect advertising: https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/spatialmath/rotation_matrix.py#L8-L17
- Our conventions should be consistent across the ecosystem - we should be using row-major.

The behavior change to both rust_utils and the python sdk are safe bc we have no users of these APIs yet. Best to fix it before we do.

Question for rust-utils maintainers: what's the release cadence here? viam-python-sdk's postinstall.sh pulls rust-utils/releases/latest, so I'll need to know when this ships as a release to time the Python SDK follow-up

### DIFF
--- a/src/ffi/spatialmath/rotation_matrix.rs
+++ b/src/ffi/spatialmath/rotation_matrix.rs
@@ -1,4 +1,5 @@
 use ffi_helpers::null_pointer_check;
+use libc::c_double;
 use nalgebra::{Matrix3, Quaternion, Rotation3, UnitQuaternion};
 
 /// The FFI interface wrapper around the nalgebra crate for RotationMatrix functions
@@ -31,8 +32,9 @@ pub unsafe extern "C" fn free_rotation_matrix_memory(ptr: *mut Rotation3<f64>) {
 }
 
 /// Initialize a 3D rotation matrix from raw components and retrieve the C pointer
-/// to its address. This function DOES NOT check whether the matrix elements provided
-/// form a valid member of SO(3)
+/// to its address. Elements are interpreted in row-major order: elements[3*r + c]
+/// is the element in row r and column c. This function DOES NOT check whether
+/// the matrix elements provided form a valid member of SO(3)
 ///
 /// # Safety
 ///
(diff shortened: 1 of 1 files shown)


src/ffi/spatialmath/rotation_matrix.rs
```diff
@@ -44,7 +46,12 @@ pub unsafe extern "C" fn viam_new_rotation_matrix(
     elements: *const [f64; 9],
 ) -> *mut Rotation3<f64> {
     null_pointer_check!(elements);
-    let matrix = Matrix3::from_vec(Vec::from(*elements));
```
JosephBorodach, line 47: https://docs.rs/nalgebra/latest/nalgebra/base/struct.Matrix.html#method.from_vec states

> Creates a matrix with its elements filled with the components provided by a slice in column-major order

nfranczak (commented):
math checks out.
i defer approval to the sdk team.
please make sure that universal-robots still remains working as it uses rust-utils through FFI

JosephBorodach:
> please make sure that universal-robots still remains working as it uses rust-utils through FFI

Yep already noted this in the rdk pr audit - universal-robots rust_utils.h only imports OV/Quaternion/Euler/AxisAngle FFI symbols, zero rotation-matrix functions. This PR only touches rotation-matrix symbols so its imports are untouched

njooma (approved):
Releases are done weekly. Sorry to have missed today's release, but if needed we can do a second release today. Or it can wait till next week. 

the python -sdk will automatically pick up a new rust-utils when we do a release

JosephBorodach:
> Releases are done weekly. Sorry to have missed today's release, but if needed we can do a second release today. Or it can wait till next week. 
> 
> the python -sdk will automatically pick up a new rust-utils when we do a release

Synced offline with @njooma, will release Python sdk before rust utils next week to give me time to make the Python sdk fix

